PR #6, opened 2020-03-20
Enforce -Wextra -Werror

ddn0:
Various fixes to bring the codebase closer to automated testing, more static analysis and more continual testing.

I moved to C++17 due to https://github.com/IntelligentSoftwareSystems/Galois/issues/69

Related:
- IntelligentSoftwareSystems/Galois#75

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,17 +86,6 @@ enable_testing()
 
 ###### Configure compiler ######
 
-#Always include debug info and aggressive warnings
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -Wall")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -Wall")
-
-# Some apps break with -O3 on recent clang versions.
-# Tracking that down is a work in progress, so for the current release,
-# only use -O2 with clang.
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND "${CMAKE_BUILD_TYPE}" STREQUAL "Release")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2")
-endif()
-
 # generate compile_commands.json
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -109,6 +98,45 @@ if(NOT "$ENV{GCC_BIN}" STREQUAL "")
...
+  # now:
+  # - Wno-deprecated-copy (boost::counting_iterator)
+  # - Wno-cast-functions (openmpi)
+  # - Wno-unused-parameter (conflicts with readability rules)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-copy -Wno-cast-function-type -Wno-unused-parameter")
+endif()
+
+# Clang
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # Some apps break with -O3 on recent clang versions.
```
tylershunt: I take it we think these are app bugs? Or is -O3 on clang known to be unstable?

libgalois/include/galois/LargeArray.h
```diff
@@ -322,9 +322,9 @@ class LargeArray<void> {
   typedef size_t size_type;
   typedef ptrdiff_t difference_type;
   typedef value_type reference;
-  typedef const value_type const_reference;
+  typedef value_type const_reference;
```
tylershunt: Seems like we should just get rid of the "const_reference" type no?

libgalois/include/galois/LargeArray.h
```diff
@@ -322,9 +322,9 @@ class LargeArray<void> {
...
   typedef value_type* pointer;
-  typedef const value_type* const_pointer;
+  typedef value_type* const_pointer;
```
tylershunt: const_pointer doesn't seem useful anymore either.

CMakeLists.txt
ddn0: Could be another instance of IntelligentSoftwareSystems/Galois#69 (failure allocating aligned memory)?

@insertinterestingnamehere do can you point me to a symptom, and I can file a ticket to retest that case?

libgalois/include/galois/LargeArray.h
ddn0: I think having all the typical container typedefs around is nice because LargeArray can be used as a drop-in replacement for an STL container. I suppose there will be a limit to this since const-ness won't propagate through return values of LargeArray, but on balance, it seems reasonable to at least implement the interface if we can. What do you think?

CMakeLists.txt
insertinterestingnamehere: I've been treating the two issues as distinct, but since unaligned accesses are UB, the could be the same. All misaligned accesses show up in clang's undefined behavior sanitizer.

libgalois/include/galois/LargeArray.h
tylershunt: I'm all for STL compatibility; my guess is that you had to remove 'const' to clear some warning. Does providing an actual 'const' reference/pointer type require a lot of effort? If so I would settle for accepting the pull request as is and opening an issue to provide real const-ness sometime in the future.

CMakeLists.txt
tylershunt: Seems like this isn't super relevant to this pull request so I don't see it as blocking.

PR #28, opened 2020-03-29
Allow architecture-specific optimizations to be skipped

### LINKED ISSUE #26: fix cmake 3.17 warning
CMake 3.17 complains about mismatching package names:
```
CMake Warning (dev) at /usr/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (ARCH_FLAGS)
  does not match the name of the calling package (ArchFlags).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/Modules/FindArchFlags.cmake:40 (find_package_handle_standard_args)
  CMakeLists.txt:164 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
Build succeeds but I haven't been able to test install. It would be nice to clear this warning eventually but I wasn't able to find a quick fix in a cursory search.

ddn0:
march=native builds optimized but non-portable binaries. When building
binaries for distribution or sharing between machines, a more careful
approach to platform optimization must be taken.

A better approach might be to provide optimization profiles that target
broadly available instruction extensions, e.g., AVX2. But there seems to
be a number of tuning parameters (e.g., cache sizes) and
compiler-specific flags involved, so I am deferring this for now.

Along the way, refactor find_packages(ArchFlags) to
include(CheckArchFlags) since determining compiler flags does not have
much similarity to finding packages. find_package_handle_standard_args
in FindArchFlags.cmake also was raising warnings due its non-standard
form:

  The package name passed to `find_package_handle_standard_args` (ARCH_FLAGS)
  does not match the name of the calling package (ArchFlags).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.

Closes #26


ddn0:
Confirmed that this resolves #26 by running cmake 3.17 on my dev branch.

PR #29, opened 2020-03-30
Perform partial file reads from s3 to support cusp

tylershunt:
I think the code is in good enough shape now to get some more eyes on it.

The major thing to point out is that I've replaced cusp's use of OfflineGraph with what I'm calling "FileViews" which are objects that represent a read only view of some range of the file. Rather than exhaustively seeking around to file locations FileViews simply map the file. There are also graph specific views OutIndexView which is a view of the entire out index list, and PartialGraphView which encapsulates views of the out index list and slices of the edge and edge data lists.

FileViews are built on top of the tsuba mmap interface which has subsumed (partially) handling local files and also handles downloads from S3.

I also made the choice to have every host download *all* of the out index list. This simplified the code somewhat, and I don't think impacts performance at all since the last host had to read the whole list anyway.


tylershunt:
I guess github reorders commits in pull requests based on date; I updated the timestamps on my commit messages to make them appear in the right order here.

ddn0:
Yeah, this is an annoying GitHub gotcha.

The professed reason from Github is that ordering by date is more in line with the conversation view, which is also in time order.

I tend to `git --amend --reset-author` the tip commit if it is not newest otherwise the CI badge goes on the wrong commit.

Some other hacky ways:
  https://stackoverflow.com/questions/44704519/how-to-fix-commit-order-in-github-pull-requests-broken-by-git-rebase

How important it is to work around is another question.

libgalois/include/galois/files/PartialGraphView.h
```diff
@@ -0,0 +1,130 @@
...
+#include <cstdio>
+#include <memory>
+
+#include <boost/iterator/counting_iterator.hpp>
+
+#include "galois/files/OutIndexView.h"
+#include "galois/files/FileView.h"
+#include "galois/graphs/GraphHelpers.h"
+
+template <typename T, typename V>
```
ddn0: put into the galois or internal/details namespace to avoid polluting the global one?

libgalois/include/galois/files/PartialGraphView.h
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  template <
+      typename EdgeType,
+      typename std::enable_if<!std::is_void<EdgeType>::value>::type* = nullptr>
+  int bind_edgeData_file() {
+    sz_t all_off = m_outidx.view_size() + sizeof(Edge_t) * m_outidx.sizeEdges();
+    /* fix the alignment since V1 can be shifted by 32 bits */
+    all_off        = align_up<uint64_t>(all_off);
+    sz_t first_off = all_off + *edge_begin() * sizeof(EdgeData_t);
```
ddn0: just confirming my understanding: `assert(*edge_begin == 0)` is always true right? seeing `base + size * sizeof(EdgeData_t)` for a starting offset confused me for a second.

libgalois/include/galois/files/PartialGraphView.h
```diff
@@ -0,0 +1,130 @@
...
+  typedef boost::counting_iterator<uint64_t> edge_iterator;
+  typedef boost::counting_iterator<uint64_t> iterator;
+
+  PartialGraphView(const PartialGraphView&) = delete;
+  PartialGraphView& operator=(const PartialGraphView&) = delete;
+
+  PartialGraphView(OutIndexView&& outidx)
+      : m_outidx(std::move(outidx)), m_filename(m_outidx.filename()) {}
+
+  PartialGraphView(PartialGraphView&& other) noexcept
```
ddn0: I always forget the specifics of moves. Is there a reason to prefer this over `PartialGraphView(...&& other) noexcept = default;` ?

libtsuba/include/tsuba/tsuba_api.h
```diff
@@ -0,0 +1,32 @@
+#ifndef _LIBTSUBA_INCLUDE_TSUBA_TSUBA_API_H_
```
ddn0: is `xyz_api.h` preferred for the C API? I had a similar question when I was prototyping the C API for `galois::do_all` and I guessed `cxyz.h`. Thoughts?

libtsuba/src/s3.cpp
```diff
@@ -1,59 +1,143 @@
...
+      if (bufferStream == nullptr) {
+        abort();
+      }
+      return bufferStream;
+    });
+    s3_client->GetObjectAsync(object_request, callback);
+  }
+
+  std::unique_lock<std::mutex> lk(m);
+  cv.wait(lk, [&] { return finished >= parts.size(); });
```
ddn0: I don't know if this matters to this API but libgalois is pretty aggressive about binding threads to cores, even statically. You might have to set the environment variable `GALOIS_DO_NOT_BIND_MAIN_THREAD=1` to get threads unpinned for async execution.

(And it might make sense to make that the default going forward.)

libtsuba/src/tsuba_api.cpp
```diff
@@ -6,31 +6,177 @@
...
+      munmap(m_ptr, m_size);
+    }
+  }
+  uint8_t* ptr() { return m_ptr; }
+};
+
+static std::mutex allocated_memory_mutex;
+static std::unordered_map<uint8_t*, MappingDesc> allocated_memory;
+
+void tsuba_init() {
```
ddn0: perhaps bubble up `tsuba_init` to the C API to force/let callers also manage tsuba shutdown/atexit ?

libtsuba/src/tsuba_api.cpp
```diff
@@ -6,31 +6,177 @@
...
+  tsuba_init();
+  MappingDesc new_mapping;
+  if (int err = new_mapping.init(filename, begin, size); err) {
+    return nullptr;
+  }
+  std::lock_guard<std::mutex> guard(allocated_memory_mutex);
+  auto [it, inserted] =
+      allocated_memory.emplace(new_mapping.ptr(), std::move(new_mapping));
+  if (!inserted) {
+    std::cerr << "Failed to emplace! (impossible?) " __FILE__ ":" << __LINE__
```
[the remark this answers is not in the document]
tylershunt: Yes, factoring out logging into a common support library seems like the right way to go about it.

libcusp/include/galois/graphs/DistributedGraph.h
```diff
@@ -33,10 +33,11 @@
 #include "galois/graphs/LC_CSR_Graph.h"
 #include "galois/graphs/BufferedGraph.h"
 #include "galois/runtime/DistStats.h"
-#include "galois/graphs/OfflineGraph.h"
 #include "galois/DynamicBitset.h"
 #include "llvm/Support/CommandLine.h"
 
+#include "galois/files/OutIndexView.h"
```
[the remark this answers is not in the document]
tylershunt: Nope! Will fix.

libgalois/include/galois/files/FileView.h
```diff
@@ -0,0 +1,55 @@
+#ifndef _LIBGALOIS_INCLUDE_GALOIS_FILES_FILE_VIEW_H_
```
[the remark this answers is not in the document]
tylershunt: No preference, I guess based on the google style guide the guard for this file would be:
#ifndef GALOIS_LIBGALOIS_GALOIS_FILES_FILE_VIEW_H_
(https://google.github.io/styleguide/cppguide.html#The__define_Guard)

I'll make my guards consisting when themselves and add an issue.

libgalois/include/galois/files/FileView.h
```diff
@@ -0,0 +1,55 @@
...
+#define _LIBGALOIS_INCLUDE_GALOIS_FILES_FILE_VIEW_H_
+
+#include <string>
+
+namespace galois {
```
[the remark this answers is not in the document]
tylershunt: Will do.

libgalois/include/galois/files/PartialGraphView.h
tylershunt: Ah, nope; this is cruft from a version of my code where PartialGraphView had some state that needed to be changed on a move.

libtsuba/include/tsuba/tsuba_api.h
tylershunt: AFAIK it's not really an established convention. I really just needed another filename since I wanted to separate the actual public library symbols and the inline wrappers that make it more convenient to call those symbols from C++. It seemed to me that the wrapper header would be included more often so I gave it the shorter name.

libtsuba/src/s3.cpp
tylershunt: That's good to know! I'll test it and see if it makes a difference.

libtsuba/src/tsuba_api.cpp
tylershunt: Yeah, that seems more standard/sane. Will fix.

libgalois/include/galois/files/OutIndexView.h
```diff
@@ -0,0 +1,139 @@
...
+typedef uint64_t edgeV2_t;
+
+/* standard types */
+typedef uint64_t sz_t; /* size_t can vary by platform (e.g., Native Client) */
+typedef uint64_t vertex_t;
+
+struct GR_header {
+  uint64_t version;
+  sz_t edgetype_size;
+  sz_t numNodes;
```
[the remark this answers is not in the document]
gurbinder533: Is there a standard convention on function names and variable names? Maybe we should pick one and stick to it.

libgalois/include/galois/files/PartialGraphView.h
```diff
@@ -0,0 +1,130 @@
...
+    return m_edge_file.bind(m_filename, first_off, last_off);
+  }
+
+  template <
+      typename EdgeType,
+      typename std::enable_if<!std::is_void<EdgeType>::value>::type* = nullptr>
+  int bind_edgeData_file() {
+    sz_t all_off = m_outidx.view_size() + sizeof(Edge_t) * m_outidx.sizeEdges();
+    /* fix the alignment since V1 can be shifted by 32 bits */
+    all_off        = align_up<uint64_t>(all_off);
```
gurbinder533: Shouldn't this alignment be dependent on Edge_t?

tylershunt: This alignment should be harmless for both v1 and v2 since both have to be aligned to 8 bytes. But I think that you're right that its worth making that explicit.

libtsuba/src/tsuba_api.cpp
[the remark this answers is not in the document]
tylershunt: Created issue #32 to track this in the greater context.

libcusp/include/galois/graphs/NewGeneric.h
```diff
@@ -216,14 +226,23 @@ class NewDistGraphGeneric : public DistGraph<NodeTy, EdgeTy> {
...
+
     galois::StatTimer graphReadTimer("GraphReading", GRNAME);
     graphReadTimer.start();
-    bufGraph.loadPartialGraph(filename, nodeBegin, nodeEnd, *edgeBegin,
-                              *edgeEnd, base_DistGraph::numGlobalNodes,
-                              base_DistGraph::numGlobalEdges);
+    uint64_t nodeBegin = base_DistGraph::gid2host[base_DistGraph::id].first;
+    uint64_t nodeEnd = base_DistGraph::gid2host[base_DistGraph::id].second;
+
+    galois::PartialGraphView<EdgeTy, edgeV1_t> pview(std::move(g));
```
gurbinder533: EdgeTy here really means EdgeDataTy. There is some confusion introduced later due to the use of both. We only need EdgeDataType (represents the type of data on the edge) and edgeV1_t (represents the data type for node IDs: uint32_t for v1 version and uint64_t for v2).

libgalois/include/galois/graphs/BufferedGraph.h
```diff
@@ -218,6 +263,29 @@ class BufferedGraph {
     assert(numBytesToLoad == 0);
   }
 
+  template <
+      typename EdgeType,
+      typename std::enable_if<!std::is_void<EdgeType>::value>::type* = nullptr>
+  void
+  loadEdgeData(const galois::PartialGraphView<EdgeDataType, uint32_t>& view) {
```
gurbinder533: Can we replace EdgeType with EdgeDataType?

gurbinder533: Can uint32_t be passed as a template parameter to support bigger graphs that need uin64_t for node IDs? This change may be required in older files as well.

libgalois/include/galois/files/OutIndexView.h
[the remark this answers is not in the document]
ddn0: If we are going Google C++ style:
  https://google.github.io/styleguide/cppguide.html#Naming

The names would be:
```
filename.cpp, other_filename.cpp filename_test.cpp
TypeName, ClassName
an.instance_variable_
a_local_variable
a.struct_field
kAConstant
AFunctionName
namespace
namespace_interal
kEnumName (more rarely, ENUM_NAME)
SCARY_MACRO (prefer constexpr etc instead)
```

There is a wide variety in the current code base. We probably should stick to the above standard for any new code.

Things we can do for existing code:
1. try to keep majority style for any existing files
2. opportunistic refactoring
3. apply `clang-tidy -fix` just for naming
4. manually refactor public interfaces

I have a slight preference for doing 3/4 and then sticking to 1. But I can also see just doing the least here for old code, i.e., just 1.

libgalois/include/galois/files/FileView.h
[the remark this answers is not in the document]
tylershunt: Issue #31

libgalois/include/galois/files/PartialGraphView.h
tylershunt: I don't think that's right, edge_begin() should return the first edge for this partial graph based on what m_firstNode is. I agree that this is confusing, I'll clean it up.

tylershunt: The index of the first edge rather, which if I understand the format is also the index of the first EdgeData enrty.

ddn0: ah yes. I forgot that these are just subgraphs.

libgalois/include/galois/graphs/BufferedGraph.h
tylershunt: I'll fix the type; I actually looked into  making the entire graph templated on edge type but ultimately decided that change deserved it's own pull request since the original BufferedGraph doesn't support V2 files.

libgalois/include/galois/files/OutIndexView.h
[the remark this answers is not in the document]
tylershunt: I'm fixing all the names in code that I added to match the style guide.

libcusp/include/galois/graphs/NewGeneric.h
```diff
@@ -174,26 +178,32 @@ class NewDistGraphGeneric : public DistGraph<NodeTy, EdgeTy> {
...
     base_DistGraph::numGlobalEdges = g.sizeEdges();
     std::vector<unsigned> dummy;
     // not actually getting masters, but getting assigned readers for nodes
     base_DistGraph::computeMasters(md, g, dummy, nodeWeight, edgeWeight);
 
+    timer.stop();
+    galois::runtime::reportStatCond_Tmax<MORE_DIST_STATS>(
+        GRNAME, "MasterDistTime", timer.get());
+
+    galois::gPrint("[", base_DistGraph::id, "] Master distribution time : ",
```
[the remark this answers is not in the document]
tylershunt: Done.

libcusp/include/galois/graphs/NewGeneric.h
tylershunt: Agree, this is the was NewGeneric was written and I was adopting the custom; I'll just go ahead and patch it now.

libgalois/include/galois/graphs/BufferedGraph.h
tylershunt: Actually I'm going to leave this the way it is for now; EdgeDataType shadows the classe's template and getting rid of this typename here breaks enable_if. This whole class should probably be cleaned up; but not here.

libgalois/include/galois/runtime/SharedMem.h
```diff
@@ -35,6 +36,7 @@ class SharedMem: public galois::substrate::SharedMem {
 
 public:
   explicit SharedMem() :  m_pa(), m_sm() {
+    TsubaInit();
```
ddn0, line 39: I'm wondering if we should put this in `galois::SharedMemSys` which is one level up from `galois::runtime::SharedMem`. `galois::runtime` is supposed to basic threading and memory allocation stuff; I don't know what the AWS dependency will add.

libgalois/include/galois/PartialGraphView.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include "galois/OutIndexView.h"
+#include "galois/FileView.h"
+#include "galois/graphs/GraphHelpers.h"
+
+namespace galois {
+
+template <typename T, typename V>
+V AlignUp(const V& val) {
+  return (val + (sizeof(T) - 1)) & sizeof(T);
```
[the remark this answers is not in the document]
tylershunt: Ah yep, good catch.

libcusp/include/galois/graphs/NewGeneric.h
```diff
@@ -347,7 +361,13 @@ class NewDistGraphGeneric : public DistGraph<NodeTy, EdgeTy> {
     base_DistGraph::initializeSpecificRanges();
 
     Tgraph_construct.stop();
-    galois::gPrint("[", base_DistGraph::id, "] Graph construction complete.\n");
+    if constexpr (MORE_DIST_STATS) {
```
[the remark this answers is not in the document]
tylershunt: I was mostly trying to get a print out so that I could read this time without collecting stat files on AWS and preserving the original semantics of the timer. I'll just set it to always print for now. That seems like the simplest solution.

tylershunt:
I also moved tsuba interactions for OutIndexView into a cpp file to keep tsuba.h out of the galois headers and avoid propagating that include directory everywhere.

libgalois/include/galois/graphs/BufferedGraph.h
gurbinder533: Sounds good.

libgalois/include/galois/graphs/BufferedGraph.h
```diff
@@ -119,6 +120,33 @@ class BufferedGraph {
...
+  /**
+   * Load the out indices (i.e. where a particular node's edges begin in the
+   * array of edges) from the file.
+   *
+   * @param graphFile loaded file for the graph
+   * @param nodeStart the first node to load
+   * @param numNodesToLoad number of nodes to load
+   */
+  void
+  loadOutIndex(const galois::PartialGraphView<EdgeDataType, uint32_t>& view) {
```
gurbinder533, line 132: Minor issue: Should we use "EdgeDataType" or "EdgeDataTy"? It is a minor thing but I think it will be good to be consistent. Galois code uses "EdgeDataTy".  I am okay with anything.

tylershunt: This is one of those consistency versus minimal change questions. I named the type EdgeDataType instead of EdgeDataTy because that is what BufferedGraph calls the type. I think It would be better to merge this as is and clean up BufferedGraph explicitly in another patch.

PR #36, opened 2020-04-03
[libdist] initialize libtsuba

tylershunt:
Initialization is needed here as well with TsubaInit out of
`galois::runtime::SharedMem`

(oops)


tylershunt:
> Perhaps we can add a test case or something to catch these types of errors?
> 
> Having a dependency on a remote server is probably too much, but perhaps there is some simple introspection routine that we could check for?

That's a good idea, I don't know what that routine would be at this stage.

I think testing in general for this code is desirable; I'm just not sure what's worth burning cycles to test yet.

Issue #32, opened 2020-03-30
Factor out logging into it's own support library


tylershunt:
It makes sense to be able to control logging via one interface. Galois has fairly standard internal logging system but it can't be called from the libraries that Galois depends on (like the new storage library).

ddn0:
More context: currently galois::substrate is the home for low-level shared functionality

ddn0:
Also there are some poorly named functions like `printOutput` which write to a log file and are in `Tracer.h`:
https://github.com/KatanaGraph/katana/blob/master/libgalois/include/galois/runtime/Tracer.h#L155

PR #54, opened 2020-05-12
[libtsuba] Allow user to set aws region via environment

tylershunt:
Quickest path to push this compile time decision to runtime. The sdk
refuses to use the default region configured by the aws cli.

I see libgalois has some environment checking boilerplate functions; We should bring 
those along when we factor out logging (as mentioned in #32).


tylershunt:
Packaging fails here:
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: cannot find -latomic

Did the environment change?

PR #66, opened 2020-06-04
Add libava directory structure

yuhc:
The new libava directory will include the following components:
- `scripts/`: currently it has a script for setting up AvA
- `specs/`: currently it has an initial CUDA API spec
- `manager/`: customized AvA manager. It's yet to be created.
- ~`ava`: the submodule pointing to the utcs-scea/ava repo~
- `CMakeLists.txt`: cmake file for building the AvA manager and (probably) static AvA guestlib.


libava/scripts/setup.sh
```diff
@@ -0,0 +1,44 @@
+#/bin/bash
+
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+echo "Installing/updating essential tools..."
+
+if [ "$(uname)" == "Darwin" ]; then
```
ddn0, line 7: nit, feel free to ignore, but I find

```bash
case "$(uname)" in
  Darwin)
    # stuff
    ;;
  Linux*)
    # stuff
    ;;
esac
```

a little easier to read and maintain.

yuhc: Updated. I also like this more :) .

yuhc:
> It would be nice to get AvA building continuously like how `.github/workflows/cpp.yaml` builds and tests Katana. I think `libava/scripts/setup.sh` can be adapted to this purpose.
> 
> I haven't thought much too deeply about this directory organization but ideally it would be good if we could get a single build system with ava and the rest of the Katana libraries. This might require some changes upstream in AvA (I haven't taken a look).

I was thinking about it on an opposite direction. Because AvA requires to compile a customized LLVM, it may be better to separate the LLVM's building/setup process (setup.sh) from the main workflow for the sake of speed. The builds and tests of libava can be like `cpp.yaml`.

But that will break the idea of automated testing, which may be a bigger bummer. If the building of the customized LLVM can be cached for different tests, then it won't be a problem. I don't know whether this is possible.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "cub"]
 	path = external/cub
 	url = https://github.com/NVlabs/cub.git
+[submodule "libava/ava"]
```
tylershunt, line 7: I think it's worth discussing whether we want ava as a submodule or just an external dependency. I think (and maybe you'll correct me) that we only need a pointer to the AVA shim library at link/run time; and then all the AVA code should just operate in the way that it usually does. At most it seems like we would want to maintain a spec here (if it's sufficiently different from the vanilla cuda spec).

Tracking it as a submodule seems like something we should avoid unless we get some big benefit; does one exist?

tylershunt:
> I was thinking about it on an opposite direction. Because AvA requires to compile a customized LLVM, it may be better to separate the LLVM's building/setup process (setup.sh) from the main workflow for the sake of speed. The builds and tests of libava can be like `cpp.yaml`.

Seems like a part of the build we could cache

.gitmodules
yuhc: I agree with you. AvA is only used to generate and compile the guestlib and API server. We only need to maintain the spec and the AvA manager in Katana.
The version of AvA being used can also be expressed in the setup script. So I don't see any big benefit using the submodule.

yuhc:
> It would be nice to get AvA building continuously like how `.github/workflows/cpp.yaml` builds and tests Katana. I think `libava/scripts/setup.sh` can be adapted to this purpose.
> 
> I haven't thought much too deeply about this directory organization but ideally it would be good if we could get a single build system with ava and the rest of the Katana libraries. This might require some changes upstream in AvA (I haven't taken a look).

Shall I integrate AvA building and test in this PR, or create a new one?

tylershunt:
Oh another point is our integration wants/desires might change when we actually start using it. Afaik we don't have any apps actually running with ava right? We might want to hold off pulling the trigger here until we do.

yuhc:
> Oh another point is our integration wants/desires might change when we actually start using it. Afaik we don't have any apps actually running with ava right? We might want to hold off pulling the trigger here until we do.

I expect to complete libava's functionality before integrating it with other Katana components. One reason is that things may keep changing; another is that I'm not familiar with cmake and the changes to the cmake files may need more discussions :/.

tylershunt:
> > Oh another point is our integration wants/desires might change when we actually start using it. Afaik we don't have any apps actually running with ava right? We might want to hold off pulling the trigger here until we do.
> 
> I expect to complete libava's functionality before integrating it with other Katana components. One reason is that things may keep changing; another is that I'm not familiar with cmake and the changes to the cmake files may need more discussions :/.

TBC I'm not against merging now; it's just a consideration

ddn0:
> Shall I integrate AvA building and test in this PR, or create a new one?

This PR seems good to merge. The sooner it is integrated the less divergence we have from master and the easier for others to contribute (if necessary).

I created an issue for AvA CI #67 . It would be good to reference it when you (or whoever) takes it on just so people have a sense of where the work-in-progress is in the codebase.

>  git modules

Looks like we don't need them so not depending on submodules seems like the right way to go. I personally don't mind either way. For me the question is more about the level of coupling of dependencies.

ddn0:
Oh. One nit: @yuhc could you squash/fixup some of your commits here? Given the final state of this PR, just having a single commit might be appropriate. Future readers looking at the commit history probably won't care too much about the intermediate states here.

tylershunt:
> Thought it would be done during the merge.

We've been preferring "Rebase and merge" w/o squashing especially for cases like this where maintaining a linear history doesn't have super high overhead.

ddn0:
> Sure. I've squashed them into a single commit. Thought it would be done during the merge.

There is an option to squash on merge in GitHub, and I'm not opposed to using that more often (it makes iterating in PRs easier to manage in GitHub), but since we are currently in a situation where sometimes people want to merge multiple commits (and we have to support upstream merges), it seems easiest to keep the state of the PR/branch exactly as one intends it to be when merged.

PR #93, opened 2020-06-24
libsupport: Add header for platform specific information

tylershunt:
I guess MAP_POPULATE is not POSIX; it breaks the build on Mac OS. There
is a whole bunch of ifdefs scattered in places like
`libgalois/src/PageAlloc.cpp` and `libgalois/src/FileGrpah.cpp` to
protect things like this. I've left those untouched because I don't have
the cycles to make sure I don't break anything.

I think having a place like Platform.h were we can keep track of these
things in a central location is a good idea. So far I've done the bare
minimum so that this will compile on Mac.


libsupport/include/galois/Platform.h
```diff
@@ -0,0 +1,28 @@
+#ifndef GALOIS_LIBSUPPORT_GALOIS_PLATFORM_H_
+#define GALOIS_LIBSUPPORT_GALOIS_PLATFORM_H_
+
+#include <sys/mman.h>
+
+#if __linux__
+#include <linux/version.h>
+#if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 22)
+#define _GALOIS_PLATFORM_MAP_POPULATE_AVAILABLE
```
ddn0, line 9: Apparently, underscore followed by capital is reserved: 

https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

Maybe just trailing underscore instead?

tylershunt: Sure

PR #99, opened 2020-06-24
Change setup to use conan/cmake integration

witchel:
Update setup script to actually install conan and set it up.  Then update instructions to use Tyler's nifty cmake integration for conan.


ddn0:
I changed the required checks... the easiest fix would be be to rebase this on top of master.

witchel:
> I changed the required checks... the easiest fix would be be to rebase this on top of master.

Thank you.  I wasn't quite sure what was going wrong with the tests...

PR #112, opened 2020-06-27
Change AWS SDK to use a custom allocator to match Conda builds.

arthurp:
Specifically the conan build now uses shared AWS-SDK since that causes AWS-SDK to use a custom allocator.

This addresses the differences between upcoming Conda builds and the local builds that were discussed in #108.

@ddn0, I know you were not involved in the previous discussion, but I have you here to review to make sure my change will not break things.


ddn0:
Seems ok to me; @tylershunt do you have any concerns?

tylershunt:
> Seems ok to me; @tylershunt do you have any concerns?

Nope! I think breaking all builds if we miss-use AWS vs std strings is the right move since it's required for conda.

PR #113, opened 2020-06-29
Patch GRPC to avoid gettid build failure

ddn0:
After #106


yuhc:
Are you able to build libava with this patch? (It only patches the gRPC in Katana, and I may also apply this patch to the gRPC in AvA as well.)

ddn0:
> Are you able to build libava with this patch

The CI builds and it builds locally. Is there something else I should test to confirm that everything is ok?

yuhc:
> The CI builds and it builds locally. Is there something else I should test to confirm that everything is ok?

AvA builds a same gRPC as libava(/manager) does. It's interesting if it works with only one gRPC patched.
Without this patch, Did libava fail to build gRPC but the cloned AvA succeed to build it?

ddn0:
My previous build directory was dirty.

Retested from a clean build and verified that the build is successful (where previously I would get the gettid issue).

PR #123, opened 2020-06-30
Create FileFrame object

scober:
FileFrame is the write path equivalent of FileView. It provides a logical abstraction of a file to the PropertyFileGraph object while transparently interacting with different forms of persistent storage through Tsuba.


libtsuba/tests/fileobjects.cpp
```diff
@@ -0,0 +1,159 @@
...
+  GALOIS_LOG_ASSERT(aro_sts.ok());
+  GALOIS_LOG_ASSERT(ff.closed());
+}
+
+int main() {
+  uint64_t num_bytes = 1 << EXP_WRITE_COUNT;
+  uint8_t bits[num_bytes];
+  fill_bits(bits, num_bytes);
+
+  auto unique_result = galois::CreateUniqueDirectory("/tmp/fileobjects-");
```
ddn0, line 149: not now, but would people enjoy (1) using a standard testing package like https://github.com/catchorg/Catch2 , (2) a RAII temporary directory utility function, and/or (3) testing macros that don't std::abort the program but instead call destructors?

FWIW, I don't mind either way, but if any of the above gets people to write more lovely unit tests like this one, it would be good to consider.

libgalois/src/PropertyFileGraph.cpp
```diff
@@ -122,38 +124,34 @@ class MMapper {
...
-    return galois::ErrorCode::InvalidArgument;
-  }
 
-  uint64_t* data = reinterpret_cast<uint64_t*>(mmapper->ptr);
-  data[0]        = 1;
-  data[1]        = 0;
-  data[2]        = num_nodes;
-  data[3]        = num_edges;
+  uint64_t data[4] = {1, 0, num_nodes, num_edges};
+  ff->Write(&data, 4 * sizeof(uint64_t));
```
ddn0, line 139: Probably return `Write` errors?

libgalois/src/PropertyFileGraph.cpp
```diff
@@ -122,38 +124,34 @@ class MMapper {
...
 
   if (num_nodes) {
-    uint64_t* out   = &data[4];
-    const auto* raw = topology.out_indices->raw_values();
-    static_assert(std::is_same_v<std::decay_t<decltype(*raw)>,
-                                 std::decay_t<decltype(*out)>>);
-    std::copy_n(raw, num_nodes, out);
+    auto buf = std::make_shared<arrow::Buffer>(
+        reinterpret_cast<const uint8_t*>(topology.out_indices->raw_values()),
+        num_nodes * sizeof(uint64_t));
```
ddn0, line 144: I kinda liked the defensive `static_assert` that was here previously to remind us to update this code if the underlying buffer type changed. The reason I put it in was that we might change `out_dests` (below) to `uint64_t` in the future.

scober: Makes sense

libgalois/src/PropertyFileGraph.cpp
scober: When you're right, you're right. And when the compiler says you're right, you're probably right.

libgalois/src/PropertyFileGraph.cpp
scober: @ddn0 Are you okay with comparing against the expected types directly?

ddn0: totally fine by me; `static_assert(sizeof(X) == sizeof(Y))` would do as well.

libgalois/src/PropertyFileGraph.cpp
```diff
@@ -122,38 +124,44 @@ class MMapper {
   size_t size{};
 };
 
-galois::Result<std::unique_ptr<MMapper>>
```
tylershunt, line 125: Could you also delete the `MMapper` class while you're in here? I think this is the only use.

libtsuba/tests/fileobjects.cpp
tylershunt: I am into Catch2; looks cool. No opinion about the others.

libgalois/src/PropertyFileGraph.cpp
ddn0: let's pour one out for `MMapper` :beers:

tylershunt: 🍻

scober: 🍻

libtsuba/include/tsuba/FileFrame.h
```diff
@@ -0,0 +1,73 @@
+#ifndef GALOIS_LIBGALOIS_GALOIS_FILE_FRAME_H_
```
ddn0, line 1: this probably should be `GALOIS_LIBTSUBA_TSUBA....`

libtsuba/tests/fileobjects.cpp
witchel: Catch2 looks cool.  No opinion about the other options.  I'm my own man though.  Tyler does not mentally control me.

PR #138, opened 2020-07-02
Add support for PropertyGraphs in Cython bfs

gurbinder533:
This ports Cython bfs to use the current property graphs interface. I have tried to work with the existing functionality provided in the PropertyFileGraph. Based on this conversion we may need to add more to the propertyFileGraph API for ease of use. 

To run new bfs:
`import galois.property_graph as PG` 
`import galois.bfsPG as bpg`
`g_ldbc = PG.PropertyGraph("/home/ubuntu/data/inputs/ldbc_003/meta")` ## replace with path to the metadata file
`bpg.bfsPG(g_ldbc,2,"p6")` ##arg: propertyGraph, sourceNode, and new propertyName

This is a good point to get the discussion started before we put more effort in porting other applications.


python/galois/_bfs_property_graph.pyx
```diff
@@ -0,0 +1,131 @@
...
+    # Append new property
+    # TODO: Remove the old property with the same name, if it exists
+    #
+    cdef shared_ptr[arrowTable] node_table = MakeTable(propertyName, distance)
+    deref(graph.get_ptr()).AddNodeProperties(node_table)
+
+#
+# Main callsite for Bfs
+#        
+def bfsPG(PropertyGraph graph, unsigned int source, str propertyName):
```
arthurp, line 123: This should just be `bfs`. No reason to give it a different name. It's in a package that makes it clear what kind of bfs it is.

python/galois/bfs_property_graph.py
```diff
@@ -0,0 +1,26 @@
...
+
+    from galois.shmem import *
+    print("Using threads:", setActiveThreads(args.threads))
+
+    bfsPG(args.graph, args.startNode, args.propertyName)
+
+    #print("Node {}: {}".format(args.reportNode, g.getData(args.reportNode)))
+
+    #if not args.noverify:
+    #    verify_bfs(g, args.startNode)
```
arthurp, line 26: Why is this stuff commented? It looks useful.

libpygalois/include/galois/Constants.h
```diff
@@ -34,4 +37,32 @@ struct ReqPushWrap {
   }
 };
 
+
+/*
+ * Extra helper functions for PropertyFileGraph
+ */
+std::shared_ptr<arrow::Table> MakeTable(const std::string& name,
+                                        const std::vector<uint32_t>& data) {
```
arthurp, line 45: Could we use the [Arrow Cython API](https://arrow.apache.org/docs/python/extending.html#cython-api) instead of this? That would avoid needing our own wrappers.

python/galois/_property_graph.pxd
```diff
@@ -0,0 +1,20 @@
+# cython: cdivision = True
```
arthurp, line 1: Can we name this module just `property_graph`? There doesn't seem to be a reason to have a python wrapper and having an internal name (with a leading underscore) appear in "application" code is bad.

python/galois/_property_graph.pyx
```diff
@@ -0,0 +1,45 @@
...
+    def num_nodes(self):
+        return self.graphTopology.num_nodes()
+
+    def num_edges(self):
+        return self.graphTopology.num_edges()
+
+    def print_node_schema(self):
+        return deref(deref(self.propertyFileGraph_ptr).node_schema()).ToString()
+
+    def get_edge_schema(self):
```
arthurp, line 28: This should be "print_edge_schema" to match above.

python/galois/_property_graph.pyx
```diff
@@ -0,0 +1,45 @@
...
+        return self.graphTopology
+
+    def num_nodes(self):
+        return self.graphTopology.num_nodes()
+
+    def num_edges(self):
+        return self.graphTopology.num_edges()
+
+    def print_node_schema(self):
+        return deref(deref(self.propertyFileGraph_ptr).node_schema()).ToString()
```
arthurp, line 26: Could we expose the Schema as an object by returning the pyarrow schema object? It should be possible via the arrow Cython API.

python/galois/cpp/libstd/boost.pxd
```diff
@@ -0,0 +1,35 @@
+from libcpp.string cimport string
+from libcpp.vector cimport vector
+from libc.stdint cimport *
+from libcpp.memory cimport shared_ptr
+cdef extern from "<boost/outcome/outcome.hpp>" namespace "outcome" nogil:
+
+    cdef cppclass std_result[T]:
+        T value()
+        pass
```
arthurp, line 9: Remove `pass`.

python/galois/cpp/libstd/boost.pxd
```diff
@@ -0,0 +1,35 @@
...
+
+cdef extern from "<arrow/api.h>" namespace "arrow" nogil:
+    cdef cppclass arrowTable "arrow::Table":
+        pass
+    cdef cppclass Array:
+        pass
+    cdef cppclass ArrayVector:
+        shared_ptr[Array] at(uint64_t)
+        uint64_t size()
+        pass
```
arthurp, line 19: Remove `pass`.

Also please add blank lines between classes (like Python coding style).

python/galois/cpp/libstd/boost.pxd
```diff
@@ -0,0 +1,35 @@
...
+from libcpp.vector cimport vector
+from libc.stdint cimport *
+from libcpp.memory cimport shared_ptr
+cdef extern from "<boost/outcome/outcome.hpp>" namespace "outcome" nogil:
+
+    cdef cppclass std_result[T]:
+        T value()
+        pass
+
+cdef extern from "<arrow/api.h>" namespace "arrow" nogil:
```
arthurp, line 11: This isn't boost, so it needs to be in it's own file.

Also can we use the Cython API for arrow instead of rolling our own? (I know I'm repeating myself, sorry, I'm not really reviewing in order)

python/galois/property_graph.py
```diff
@@ -0,0 +1,10 @@
+from ._property_graph import PropertyGraph
```
arthurp, line 1: I don't think this module serves any purpose. Being able to create and immediately discard a PropertyGraph from the command line doesn't really do much. And as I said before I want to avoid having `_property_graph` appear in user Cython code.

python/galois/shmem.pxd
```diff
@@ -3,18 +3,22 @@
...
+#from .cpp.libstd.boost cimport std_result
+from .cpp.libstd.boost cimport *
 from libcpp.vector cimport vector
 from libcpp.string cimport string
 from libcpp cimport bool
 import sys
 from libc.stdint cimport *
 from libc.math cimport fabs
+from ._property_graph cimport PropertyGraph
+from libcpp.memory cimport shared_ptr, static_pointer_cast
```
arthurp, line 21: Let's try to avoid reexporting everything from shmem. It doesn't make much sense because shmem should just contain the runtime engine, not all the data structures and utilities. If we want to have a universal import that picks up everything let's make it the `galois` package itself (using a file `__init__.pxd` which I assume is supported). But honestly I'd prefer to have the program import property graphs and whatever else only if they need it.

libpygalois/include/galois/Constants.h
```diff
@@ -1,3 +1,6 @@
+#include "galois/graphs/PropertyFileGraph.h"
```
arthurp, line 1: Could libpygalois be replaced with verbatim C++ code in the Cython files? It is supported: https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#including-verbatim-c-code

python/galois/_bfs_property_graph.pyx
```diff
@@ -0,0 +1,131 @@
...
+    cdef:
+        atomuint32_t notVisited
+        GReduceMax[uint32_t] maxDist
+        uint32_t source = <uint32_t> source_i
+        shared_ptr[UInt32Array] chunk
+        uint32_t chunk_offset = 0
+        uint64_t numNodes = graph.num_nodes()
+        uint32_t start, end
+
+
```
arthurp, line 40: Clean up white space.

python/galois/_bfs_property_graph.pyx
```diff
@@ -0,0 +1,131 @@
+# cython: cdivision = True
+
+from galois.shmem cimport *
+from cython.operator cimport dereference as deref
+
+ctypedef atomic[uint32_t] atomuint32_t
```
arthurp, line 6: atomuint32_t doesn't seem any shorter than atomic[uint32_t]. Maybe just use the latter everywhere. Less to look up for a reader.

python/galois/_bfs_property_graph.pyx
```diff
@@ -0,0 +1,131 @@
...
+        if(n == source):
+            distance[n] = 0
+        else:
+            distance[n] = numNodes
+
+cdef void not_visited_operator(uint64_t numNodes, atomuint32_t *notVisited, shared_ptr[UInt32Array] data, uint32_t n):
+    cdef:
+        uint32_t val
+    val = deref(data).Value(n)
+    if (val >= numNodes):
```
arthurp, line 20: No parens.

python/galois/_property_graph.pxd
```diff
@@ -0,0 +1,20 @@
...
+from galois.shmem import *
+from libcpp.memory cimport shared_ptr
+from cython.operator cimport dereference
+
+#
+# Python Property graph 
+#
+cdef class PropertyGraph:
+    cdef:
+        shared_ptr[PropertyFileGraph] propertyFileGraph_ptr
```
arthurp, line 13: Since we own this graph object, could be just use a normal pointer? It would save a lot of `deref` calls.

python/galois/_bfs_property_graph.pyx
gurbinder533: Done

python/galois/bfs_property_graph.py
gurbinder533: Done

libpygalois/include/galois/Constants.h
gurbinder533: Involves a bigger change. Will save it for the future.

python/galois/_property_graph.pxd
gurbinder533: Done

python/galois/_property_graph.pyx
gurbinder533: Good catch.

libpygalois/include/galois/Constants.h
gurbinder533: Can be done later.

python/galois/_bfs_property_graph.pyx
gurbinder533: Done

python/galois/property_graph.py
```diff
@@ -0,0 +1,10 @@
+from .property_graph import PropertyGraph
```
arthurp, line 1: Having a .py and .pyx with the same name will almost certainly cause problems. Just remove this file. I don't think it does anything as it stands.

python/galois/cpp/libstd/boost.pxd
arthurp: We can drop all my other code style stuff, but this one I'm gonna stick to because I don't want to have to find this Arrow related code in a file named `boost` when we need to change/remove it.

gurbinder533:
Good point about arrow and boost. That was supposed to be a temporary measure to get things running fast :P. They are separate now.

PR #170, opened 2020-07-07
Enable ccache for conda builds

arthurp:
The builds should work fine in environments without ccache installed if needed and do not install ccache inside conda. They check for ccache and use it if it is available.


arthurp:
I thought this was working. It's not. It's working for me on my local machine. I'm not sure what the difference is. I'll continue investigating.

arthurp:
I forgot to install ccache in the build environment. It's working now. One example is: first run 28m then after rebasing the PR 15m.

PR #175, opened 2020-07-08
Continually test documentation

ddn0:


.github/workflows/setup_ubuntu.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-apt install -yq ccache curl gcc-9 g++-9 libopenmpi-dev
+apt install -yq ccache curl doxygen gcc-9 g++-9 libopenmpi-dev
```
witchel, line 2: My suggestion would be  doxygen doxygen-doc graphviz graphviz-doc

ddn0: graphviz makes sense. What's the purpose of the `-doc` packages?

witchel: Yeah, I thought you would ask about that.  For ci they would be pointless.  But for a developer... I dunno.  Isn't, like, being an autodidact a good thing?

witchel: Oh, sorry, I thought this was `setup_dev_ubuntu.sh`.  That one should probably be updated too.

ddn0: Ah sure. Updated setup_dev_ubuntu.sh.

PR #179, opened 2020-07-08
Handle neo4j import data with more property types efficiently

patrickkenney9801:
This PR handles the issue of a limit of 2GB of data per arrow array by using chunked arrays and outputs parquet files with multiple RowGroups
This PR also better handles adding null values by doing so in bulk instead of in step with the building of the in-memory dataset.
This gives significantly better performance for datasets with many properties, especially if related objects in the dataset have high locality.
This PR does not solve handling datasets that cannot fit in RAM as arrow arrays.


tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -149,6 +149,29 @@ void addFalseColumnBuilder(const std::string& column,
...
+// not null
+void addFalseColumnBuilder(const std::string& column,
+                           std::unordered_map<std::string, size_t>* map,
+                           BooleanBuilders* columnBuilders,
+                           ArrowFields* schemaVector,
+                           std::shared_ptr<arrow::DataType> type,
+                           size_t offset) {
+  // add entry to map
+  map->insert(std::pair<std::string, size_t>(column, map->size()));
+  auto entry = map->find(column);
```
tylershunt, line 162: Nit picking; but `map->insert()` returns the iterator that you're looking up again here. you could do something like:
```C++
auto [entry, found] = map->insert(....)
assert(found);
```

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -946,6 +967,29 @@ void appendValue(std::shared_ptr<arrow::ArrayBuilder> array,
   }
 }
 
+void evenOutTable(ArrayBuilders& arrays, size_t currentElts) {
+  arrow::Status st = arrow::Status::OK();
+  for (auto array : arrays) {
+    uint64_t nullCount = currentElts - array->length();
+    st                 = array->AppendNulls(nullCount);
```
tylershunt, line 974: I don't have good context here but arrays here are columns right? And columns are indexed by node id? If so it seems weird to me that the nulls would always come at the end.

patrickkenney9801: In this case, this occurs if a node does not have a certain property, that is the property is null.  So I append these nulls at the end to make sure every row has each column, whether it is null or valid data

tylershunt: But this assumes that there won't be a node without this property in the middle (or you handle that case somewhere else)? Or are the arrays for different properies indexed differently?

patrickkenney9801: This is only called once all nodes have been read in, to even out the table by ensuring all columns are of the same length, since if a column is not even then its missing entries are all null by construction

tylershunt: Cool cool; I think I'm getting it now.

ddn0: this seems like a useful invariant to sanity check: that the length of a node property matches the number of nodes and similarly for edge properties.

patrickkenney9801: this is the sanity check essentially--

ddn0: yes. But this should be true of any final output regardless of input format. So I'm suggesting defensively pulling up this invariant as precondition to writing out a property graph file.

patrickkenney9801: good idea, but I think, not sure if this is true, that all arrow::Tables must have the same number of rows, in which case since we pass PFG tables, it is handled that way

tylershunt: Yeah but we could be sure if we checked 😛.

libtsuba/src/RDG.cpp
```diff
@@ -223,9 +224,9 @@ WriteTable(const arrow::Table& table,
       return res.error();
     }
 
-    auto write_result =
-        parquet::arrow::WriteTable(*column, arrow::default_memory_pool(), ff,
-                                   std::numeric_limits<int64_t>::max());
+    auto write_result = parquet::arrow::WriteTable(
+        *column, arrow::default_memory_pool(), ff, kRowGroupSize);
+    // std::numeric_limits<int64_t>::max());
```
tylershunt, line 229: Would rather just ditch this comment or move it up to the constant.

ddn0:
@patrickkenney9801 is there a test input or generator that we can add that would test some aspect of this new functionality? E.g., an input with nulls, configuring row size manually (to some small number) and observing that the output data is the same with row size 1 and with row size N, ...?

patrickkenney9801:
@ddn0 I was considering making row size a user parameter so I will just do that and have it default to 25K, make a test from the existing files with size 3 or so.

tools/graph-convert/graph-properties-convert.cpp
nicelhc13: Seems like this file is outdated, but let me ask a question on this since I can still see AppendNull(). Could we use `galois::DynamicBitset` instead of the dense bool array? It would add more latency overhead but the memory overhead is the main issue.

patrickkenney9801: That is possible, but I do not think this instance of the code is very focused on memory optimization, more speed.  Since that is the case, with the required complexity and since boolean arrays only account for a small market share of memory usually.  I think this optimization should be deferred until we get more serious about memory footprint.

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -39,6 +39,7 @@ using galois::ImportDataType;
 typedef std::vector<std::shared_ptr<arrow::ArrayBuilder>> ArrayBuilders;
```
l-hoang, line 39: "using" is the C++ typedef to my understanding

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -63,34 +64,36 @@ struct KeyGraphML {
...
-    if (!st.ok()) {
-      GALOIS_LOG_FATAL("Error building arrow array");
-    }
-    arrays.push_back(array);
+/************************************/
+/* Shared GraphML/CSV/JSON Function */
+/************************************/
+
+template <typename T>
+void computePrefixSum(std::vector<T>* array) {
```
l-hoang, line 72: There's a parallel prefix sum in Galois, though this only matters if you're using threads in this converter.
https://github.com/KatanaGraph/katana/blob/675f38f103b5da165af0c7f8729b7030fb1839e0/libgalois/include/galois/ParallelSTL.h#L314

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -124,6 +127,85 @@ buildTable(std::vector<std::shared_ptr<T>>* columnBuilders,
...
+      continue;
+    }
+    st = builder->Append(array->Value(index));
+    if (!st.ok()) {
+      GALOIS_LOG_FATAL("Error appending value to an arrow array builder");
+    }
+  }
+}
+
+void rearrangeArray(std::shared_ptr<arrow::StringBuilder> builder,
```
l-hoang, line 155: This function  looks exactly the same as the above function except for "GetView" rather than "Value". I believe GetView exists for all arrow arrays as a matter of backward compatibility (might be wrong though), and it does the exact same thing as Value.

libtsuba/src/RDG.cpp
```diff
@@ -223,9 +224,8 @@ WriteTable(const arrow::Table& table,
       return res.error();
     }
 
-    auto write_result =
-        parquet::arrow::WriteTable(*column, arrow::default_memory_pool(), ff,
-                                   std::numeric_limits<int64_t>::max());
+    auto write_result = parquet::arrow::WriteTable(
+        *column, arrow::default_memory_pool(), ff, kRowGroupSize);
```
ddn0, line 228: I've been poking around the `WriteTable` code and I believe the chunk size refers to the size of a data page within a row group: https://github.com/apache/arrow/blob/apache-arrow-0.17.1/cpp/src/parquet/arrow/writer.cc#L670

tylershunt: Does that mean this is going to increase the file size?

ddn0: Yes (marginally) and in my one test (ldbc_100) it also increases loading time from the local filesystem by 3X.

I have some changes that will fix this.

PR #181, opened 2020-07-08
Caching deb packages

ddn0:
Avoid transient "mirror sync" errors by keeping cache of debs around.


ddn0:
The configuration cruft the support this for real seems to outweigh the small annoyance of infrequent failed updates.

PR #188, opened 2020-07-10
Tsuba: Use result for `file*` interfaces

tylershunt:
Added result types to the `File*` interfaces. No functional changes (hopefully).

After #178


libtsuba/include/tsuba/file.h
```diff
@@ -32,47 +32,53 @@ struct StatBuf {
 };
 
 // Download a file and open it
-int FileOpen(const std::string& uri);
+//[[deprecated]] (turning this on breaks the build)
```
ddn0, line 35: We might consider `-Wno-deprecated-declarations` in some fashion otherwise we won't be able to use `[[deprecated]]` at all ourselves.

Other option: file a ticket to remove `FileOpen` throughout the codebase and lean towards never using `[[deprecated]]` internally.

libtsuba/src/RDG.cpp
```diff
@@ -19,20 +20,54 @@
...
+  uint32_t version;
+  uint32_t num_hosts;
+
+  // NOLINTNEXTLINE needed non-const ref for nlohmann compat
+  friend void to_json(json& j, const RDGMeta& meta) {
+    j = json{{"magic", kRDGMagicNo},
+             {"version", meta.version},
+             {"num_hosts", meta.num_hosts}};
+  }
+  // NOLINTNEXTLINE needed non-const ref for nlohmann compat
```
witchel, line 51: Now that is readable.

libtsuba/src/RDG.cpp
```diff
@@ -19,20 +20,54 @@
...
+  friend void from_json(const json& j, RDGMeta& meta) {
+    uint32_t magic;
+    j.at("magic").get_to(magic);
+    j.at("version").get_to(meta.version);
+    j.at("num_hosts").get_to(meta.num_hosts);
+    if (magic != kRDGMagicNo) {
+      // nlohmann::json reports errors using exceptions
+      throw std::runtime_error("RDG Magic number mismatch");
+    }
```
witchel, line 60: That seems ok to me, but I haven't come across many throw's in our code.  I don't know if we have some sort of policy. Oh, you have a top-level catch, which makes this easier to swallow.

libtsuba/src/file.cpp
```diff
@@ -187,17 +188,17 @@ int S3OrWriteFile(const std::string& uri, const uint8_t* data, uint64_t size) {
 } // namespace
 
 // Recommend rename to FileDownloadOpenUnlink
-int tsuba::FileOpen(const std::string& uri) {
+[[deprecated]] galois::Result<int> tsuba::FileOpen(const std::string& uri) {
   if (!tsuba::IsS3URI(uri)) {
```
witchel, line 192: Shouldn't we just make this an empty function?  Right now, it can do a lot of work that effectively gets thrown away.

libsupport/include/galois/Logging.h
```diff
@@ -122,4 +123,13 @@ void LogLine(LogLevel level, const char* file_name, int line_no, F fmt_string,
...
+#define GALOIS_WARN_ONCE(fmt_string, ...)                                      \
+  do {                                                                         \
+    static std::once_flag __galois_warn_once_flag;                             \
+    std::call_once(__galois_warn_once_flag, []() {                             \
+      ::galois::LogLine(::galois::LogLevel::Warning, __FILE__, __LINE__,       \
+                        FMT_STRING(fmt_string), ##__VA_ARGS__);                \
+    });                                                                        \
+  } while (0)
+
 #endif
```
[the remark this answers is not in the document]
tylershunt: It was committed earlier today! This is cruft from that that I need to rebase out.

libtsuba/src/RDG.cpp
ddn0: I've been trying to avoid `throw`s because I don't know if all our code is exception safe. If it is possible to keep the exception scope small, which I think it is here, it is probably ok.

libtsuba/src/file.cpp
tylershunt: It's still used [here](https://github.com/KatanaGraph/katana/blob/fa4e369ab06b10b9d6fce04ee45ba717b33d1acb/libgalois/src/FileGraph.cpp#L305) where `FileGraph` expects a file descriptor. The change to make `FileGraph` use `FileMmap` is probably not huge but I wanted to keep it out of this interface only PR.

libtsuba/src/RDG.cpp
tylershunt: Yeah I went back and forth but failing at the `parse` step which is required to read data seemed more maintainable than remembering to do this check elsewhere. Unfortunately `lohmann/json` is just one of those libraries that expect exceptions so this is the best I could come up with.

libgalois/src/FileGraph.cpp
```diff
@@ -300,9 +301,13 @@ void* FileGraph::fromArrays(uint64_t* out_idx, uint64_t num_nodes, void* outs,
...
-  int fd;
+  int fd = -1;
   if (tsuba::IsS3URI(filename)) {
-    fd = tsuba::FileOpen(filename);
+    auto fd_res = tsuba::FileOpen(filename);
+    if (!fd_res) {
+      GALOIS_LOG_FATAL("tsuba opening {}: {}", filename, fd_res.error());
+    }
+    fd = fd_res.value();
   } else {
```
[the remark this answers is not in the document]
tylershunt: Tangled with supporting s3 reads for `FileGraph` see my comment below.

libgalois/include/galois/PartialGraphView.h
```diff
@@ -8,8 +8,10 @@
 #include <boost/iterator/counting_iterator.hpp>
 
 #include "galois/OutIndexView.h"
+#include "galois/Result.h"
 #include "galois/graphs/GraphHelpers.h"
 #include "tsuba/FileView.h"
+#include "galois/Result.h"
```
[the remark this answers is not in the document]
tylershunt: sometimes you have to be *extra* sure you've got it.

libtsuba/src/RDG.cpp
tylershunt: Also lohmann will throw an exception for us *anyway* if parsing fails so the catch is required regardless. The library is really nice though so I'm willing to suffer this.

libtsuba/include/tsuba/file.h
tylershunt: I think I'll opt for the latter. There's only one call site so it should be an easy fix.

tylershunt: #193

PR #189, opened 2020-07-10
docs: Add note to contributing about how to add external deps.

tylershunt:
So that my pain doesn't have to be your pain.


CONTRIBUTING.md
```diff
@@ -176,6 +176,37 @@ fail if the tests fail. These commands are primarily useful for debugging Conda
...
+  [conda-forge](https://anaconda.org/conda-forge/repo) and in
+  [ConanCenter](https://conan.io/center/) if it is not available in both places,
+  ubuntu package managers like `apt` or `snap` can work but adding it
+  will be different (and you should consider picking another library since
+  this puts an extra burden on developers).
+
+  1. Add the dependency to the [conan config](config/conanfile.txt) in the style
+     of the dependendices that are already there.
+
+  2. Add the dependency to the [conda recepe](conda_recipe/meta.yaml) in the
```
ddn0, line 196: recipe

CONTRIBUTING.md
```diff
@@ -176,6 +176,37 @@ fail if the tests fail. These commands are primarily useful for debugging Conda
...
+  style of what's there.
+
+  3. Modify the [package config](cmake/GaloisConfig.cmake.in) so that it will
+  be found during the conda build (again the best advice is to look at how other
+  dependencies handle this).
```
arthurp, line 201: This is only needed if the new dependency is a runtime or user-code build time dependency. So, for instance, header only libraries that are not used in our headers, don't need to appear here.

CONTRIBUTING.md
```diff
@@ -176,6 +176,37 @@ fail if the tests fail. These commands are primarily useful for debugging Conda
...
+
+If you do end up choosing a library that is not in conda-forge/ConanCenter
+(really?) make sure to update the dependency list in [README.md](README.md), and
+make sure the
+[script for setting up a dev environment](scripts/setup_dev_ubuntu.sh) is
+updated as well. There will likely also be changes to the CI scripts that are
+needed.
```
arthurp, line 208: Maybe add that any external deps that are not in conda-forge should be discussed with the conda package maintainer (so me at the moment). This is because anything we cannot pull in as a normal conda dep will need really careful handling, and not handling it will totally break the conda packages.

CONTRIBUTING.md
```diff
@@ -176,6 +176,37 @@ fail if the tests fail. These commands are primarily useful for debugging Conda
...
+  [ConanCenter](https://conan.io/center/). If it is not available in both places,
+  ubuntu package managers like `apt` or `snap` can work but adding it will be
+  different (and you should consider picking another library since this puts an
+  extra burden on developers).
+
+  1. Add the dependency to the [conan config](config/conanfile.txt) in the style
+     of the dependendices that are already there.
+
+  2. Add the dependency to the [conda recipe](conda_recipe/meta.yaml) in the
+  style of what's there.
```
arthurp, line 197: There are two sections: "host" and "run". Any runtime deps need to be added to both sections. But deps which are totally compiled into galois (e.i., they are not exposed in our API and don't require a shared library at run time), can be in "host" only. This is a slightly rare case, but may be worth mentioning.

CONTRIBUTING.md
tylershunt: 👍 I will soften that.

CONTRIBUTING.md
tylershunt: 👍

CONTRIBUTING.md
tylershunt: 👍 will do; I'll write down your physical address.

tylershunt:
I believe that should wrap it up.

PR #192, opened 2020-07-10
Cypher: return support for nodes

l-hoang:
This PR adds return support for nodes in cypher queries: a user can return an entire node or certain properties of nodes. **There is currently no support for returning paths or edges.** This is because we currently do not have the matched path/edge info compiled after matching; some post-processing will be required and will be implemented later. **There is also currently no support for using node properties during querying itself.** This is only *return* support.

Some examples of what this PR lets you do:

MATCH (a:Comment)-[:HAS_TAG]->(c) RETURN c

```
================================================================================
c
================================================================================
(:Tag {id: 4, url: http://dbpedia.org/resource/Humayun, name: Humayun,  } )
(:Tag {id: 59,  url: http://dbpedia.org/resource/Rupert_Murdoch,  name: Rupert_Murdoch, } )
(:Tag {id: 132, url: http://dbpedia.org/resource/Judy_Davis,  name: Judy_Davis, } )
(:Tag {id: 132, url: http://dbpedia.org/resource/Judy_Davis,  name: Judy_Davis, } )
(:Tag {id: 132, url: http://dbpedia.org/resource/Judy_Davis,  name: Judy_Davis, } )
(:Tag {id: 132, url: http://dbpedia.org/resource/Judy_Davis,  name: Judy_Davis, } )
(:Tag {id: 139, url: http://dbpedia.org/resource/Wolfgang_Amadeus_Mozart, name: Wolfgang_Amadeus_Mozart,  } )
(:Tag {id: 139, url: http://dbpedia.org/resource/Wolfgang_Amadeus_Mozart, name: Wolfgang_Amadeus_Mozart,  } )
(:Tag {id: 139, url: http://dbpedia.org/resource/Wolfgang_Amadeus_Mozart, name: Wolfgang_Amadeus_Mozart,  } )
(:Tag {id: 139, url: http://dbpedia.org/resource/Wolfgang_Amadeus_Mozart, name: Wolfgang_Amadeus_Mozart,  } )
(:Tag {id: 140, url: http://dbpedia.org/resource/Arnold_Schwarzenegger, name: Arnold_Schwarzenegger,  } )
(:Tag {id: 140, url: http://dbpedia.org/resource/Arnold_Schwarzenegger, name: Arnold_Schwarzenegger,  } )
(:Tag {id: 144, url: http://dbpedia.org/resource/Franz_Joseph_I_of_Austria, name: Franz_Joseph_I_of_Austria,  } )
(:Tag {id: 144, url: http://dbpedia.org/resource/Franz_Joseph_I_of_Austria, name: Franz_Joseph_I_of_Austria,  } )
(:Tag {id: 196, url: http://dbpedia.org/resource/Max_Mirnyi,  name: Max_Mirnyi, } )
(:Tag {id: 205, url: http://dbpedia.org/resource/Charles_V,_Holy_Roman_Emperor, name: Charles_V,_Holy_Roman_Emperor,  } )
(:Tag {id: 284, url: http://dbpedia.org/resource/Avril_Lavigne, name: Avril_Lavigne,  } )
(:Tag {id: 284, url: http://dbpedia.org/resource/Avril_Lavigne, name: Avril_Lavigne,  } )
(:Tag {id: 284, url: http://dbpedia.org/resource/Avril_Lavigne, name: Avril_Lavigne,  } )
(:Tag {id: 285, url: http://dbpedia.org/resource/Nelly_Furtado, name: Nelly_Furtado,  } )
(:Tag {id: 292, url: http://dbpedia.org/resource/Paul_Martin, name: Paul_Martin,  } )
(:Tag {id: 413, url: http://dbpedia.org/resource/Paul_Capdeville, name: Paul_Capdeville,  } )
(:Tag {id: 543, url: http://dbpedia.org/resource/Gamal_Abdel_Nasser,  name: Gamal_Abdel_Nasser, } )
(:Tag {id: 571, url: http://dbpedia.org/resource/Henry_IV_of_France,  name: Henry_IV_of_France, } )
(:Tag {id: 572, url: http://dbpedia.org/resource/Maurice_Ravel, name: Maurice_Ravel,  } )
(:Tag {id: 572, url: http://dbpedia.org/resource/Maurice_Ravel, name: Maurice_Ravel,  } )
(:Tag {id: 795, url: http://dbpedia.org/resource/George_I_of_Great_Britain, name: George_I_of_Great_Britain,  } )
(:Tag {id: 799, url: http://dbpedia.org/resource/Peter_Paul_Rubens, name: Peter_Paul_Rubens,  } )
(:Tag {id: 1025,  url: http://dbpedia.org/resource/Cliff_Richard, name: Cliff_Richard,  } )
(:Tag {id: 1167,  url: http://dbpedia.org/resource/Ehud_Olmert, name: Ehud_Olmert,  } )
(:Tag {id: 1204,  url: http://dbpedia.org/resource/Titian,  name: Titian, } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1410,  url: http://dbpedia.org/resource/Muammar_Gaddafi, name: Muammar_Gaddafi,  } )
(:Tag {id: 1452,  url: http://dbpedia.org/resource/Vincent_van_Gogh,  name: Vincent_van_Gogh, } )
(:Tag {id: 1452,  url: http://dbpedia.org/resource/Vincent_van_Gogh,  name: Vincent_van_Gogh, } )
(:Tag {id: 1506,  url: http://dbpedia.org/resource/Manmohan_Singh,  name: Manmohan_Singh, } )
(:Tag {id: 1524,  url: http://dbpedia.org/resource/Mariano_Rivera,  name: Mariano_Rivera, } )
(:Tag {id: 1524,  url: http://dbpedia.org/resource/Mariano_Rivera,  name: Mariano_Rivera, } )
(:Tag {id: 1532,  url: http://dbpedia.org/resource/Gloria_Macapagal-Arroyo, name: Gloria_Macapagal-Arroyo,  } )
(:Tag {id: 1614,  url: http://dbpedia.org/resource/Oscar_Wilde, name: Oscar_Wilde,  } )
(:Tag {id: 1687,  url: http://dbpedia.org/resource/Paul_I_of_Russia,  name: Paul_I_of_Russia, } )
(:Tag {id: 1687,  url: http://dbpedia.org/resource/Paul_I_of_Russia,  name: Paul_I_of_Russia, } )
(:Tag {id: 1753,  url: http://dbpedia.org/resource/Ban_Ki-moon, name: Ban_Ki-moon,  } )
(:Tag {id: 1753,  url: http://dbpedia.org/resource/Ban_Ki-moon, name: Ban_Ki-moon,  } )
(:Tag {id: 1759,  url: http://dbpedia.org/resource/Francisco_Franco,  name: Francisco_Franco, } )
......
```

MATCH (a:Comment)-[:HAS_TAG]->(c) RETURN count(a) as commentcount;
```
================================================================================
commentcount
================================================================================
279
================================================================================
```

MATCH (a:Comment)-[:HAS_TAG]->(c) RETURN count(a) as commentcount, c.id;
```
================================================================================
commentcount
================================================================================
279
================================================================================

================================================================================
c.id
================================================================================
4
59
132
132
132
132
139
139
139
139
140
140
144
144
196
205
284
284
284
285
292
413
```

Note that only strings and int64 properties are supported for printing; this should be enough for our purposes. List support and others will be added over time.


libquery/include/querying/IndexableChunkedArray.h
```diff
@@ -63,10 +63,12 @@ class IndexableChunkedArray {
    * @tparam NumType arrow DataType class representing a numeric type to cast
    * the array into
    * @param global_index Index of array to get
-   * @returns Element located at specified index
+   * @returns Pair saying if the element at some index is valid +
+   * element located at specified index if it is valid
    */
   template <typename NumType>
-  typename NumType::c_type getNumericElement(int64_t global_index) const {
+  std::pair<bool, typename NumType::c_type>
```
ddn0, line 70: https://en.cppreference.com/w/cpp/utility/optional ?

libquery/include/querying/IndexableChunkedArray.h
```diff
@@ -81,11 +83,17 @@ class IndexableChunkedArray {
...
+    if (numeric_array->IsValid(access_location.second)) {
+      return std::make_pair(true, numeric_array->Value(access_location.second));
+    } else {
+      return std::make_pair(false, 0);
+    }
   }
 
-  //! Intrepret this array as a string array and return a string at some index.
-  std::string getStringElement(int64_t global_index) const;
+  //! Intrepret this array as a string array and return a string at some index
```
ddn0, line 93: "Interpret"

FWIW, I'd stick with one doxygen comment style in a file. There seems to be some switching back and forth between `/*** ` and `//! `

libquery/src/AttributedGraph.cpp
```diff
@@ -302,6 +300,18 @@ AttributedGraph::getEdgeLabelMask(const std::string& edgeLabel) {
   }
 }
 
+std::string AttributedGraph::getNodeLabelString(uint32_t node_index) {
+  std::string built_string = "";
+  QueryNode& node          = this->graph.getData(node_index);
```
ddn0, line 305: `this->` seems a little superfluous but maybe that's just me.

libquery/src/AttributedGraph.cpp
```diff
@@ -302,6 +300,18 @@ AttributedGraph::getEdgeLabelMask(const std::string& edgeLabel) {
...
+  // TODO not the most efficient algorithm probably
+  for (auto& string_bit_pair : this->nodeLabelIDs) {
+    if (1u << string_bit_pair.second & node.label) {
```
ddn0, line 308: I'd add parens for clarity.

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+  std::unordered_map<std::string, uint32_t> var_name_to_node_index;
+  for (unsigned i = 0; i < node_index_to_var_name.size(); i++) {
+    std::string key = node_index_to_var_name[i];
+    if (key != "") {
+      auto emplace_result = var_name_to_node_index.try_emplace(key, i);
+      // issue if already existed; it means var name conflict
+      if (!emplace_result.second) {
+        GALOIS_LOG_FATAL("Key {} already existed var to node index map", key);
+      }
+      galois::gDebug("Key ", key, " ", i);
```
ddn0, line 405: If you are going to use the new libsupport macros, `GALOIS_LOG_DEBUG` is a replacement for `galois::gDebug`

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+          // not a count + has a property name; get the property type
+          new_metadata.all_props = false;
+          new_metadata.d_type =
+              graph.getNodeAttributeType(cur_result.property_name);
+        }
+      }
+      node_data_to_return.emplace_back(new_metadata);
+    } else {
+      // TODO would have to check edges and paths as well, not just nodes
+      galois::gWarn("Asking to return id ", cur_result.base_name,
```
ddn0, line 438: ditto

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+  return node_data_to_return;
+}
+
+/**
+ * Given a ChunkedArray and its type, print the element located at some index
+ * with some specified header/footer
+ */
+void printArrayElement(const std::shared_ptr<arrow::DataType>& d_type,
+                       const IndexableChunkedArray& prop_array, int64_t index,
+                       std::string header = "", std::string footer = "") {
```
ddn0, line 452: FWIW, I'd generally avoid function default arguments in new code.

Some reasons why: https://quuxplusone.github.io/blog/2020/04/18/default-function-arguments-are-the-devil/

And a simple function overload is clearer and accomplishes the same thing for the most part.

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+  std::string divider("========================================"
+                      "========================================\n");
+
+  for (const QueryNodeReturnMetadata& metadata : node_data_to_return) {
+    galois::gPrint(divider);
+    galois::gPrint(metadata.compiler_info.return_name, "\n");
+    galois::gPrint(divider);
+
+    if (!metadata.compiler_info.is_count) {
+      // get the embedding id we are interested int
```
ddn0, line 486: "in"

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+      nodeContains.resize(numQueryNodes, "");
+      prefixSum.resize(numQueryNodes, 0);
+      node_index_to_var_name.resize(numQueryNodes, "");
+
+      // node types check: save node types for each id
+      // NOTE: This saves the FIRST type encountered by the source ID
+      if (nodeTypes[srcID] == NULL) {
+        nodeTypes[srcID] = queryEdges[j].caused_by.labels;
+      } else {
+        // assert(std::string(nodeTypes[srcID]) ==
```
ddn0, line 596: ?

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+      std::string s2 = std::string(filters[2 * j + 1]);
+
+      // node contains check; save string filters for each node
+      // NOTE: saves first constraint encounted by some node but no otherse
+      if (nodeContains[srcID] == "") {
+        nodeContains[srcID] = s1;
+      } else {
+        assert(nodeContains[srcID] == s1);
+      }
+      if (nodeContains[dstID] == "") {
```
ddn0, line 617: `empty()` maybe

libquery/src/AttributedGraph.cpp
```diff
@@ -363,130 +373,324 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+        nodeContains[dstID] = s2;
+      } else {
+        assert(nodeContains[dstID] == s2);
+      }
+
+      // var binding for node
+      if (node_index_to_var_name[srcID] == "") {
+        node_index_to_var_name[srcID] = queryEdges[j].caused_by.var_name;
+      } else {
+        assert(node_index_to_var_name[srcID] ==
```
ddn0, line 627: style nit: I'd just pull up these assertions one level. Invariants are easiest to parse when they are unconditionally held, and although it is plain to the equivalence in the current code, one point of assertions is to be defensive in the face of future change.

libquery/include/querying/IndexableChunkedArray.h
l-hoang: Thanks for mentioning this; forgot this existed, will change.

libquery/include/querying/IndexableChunkedArray.h
l-hoang: Makes sense; just a habit of mine to use //! for what I *think* will be one-liners but what ends up happening is that they become more than one line and I don't switch to /**.

I'll switch all of the doxygen to something consistent.

libquery/src/AttributedGraph.cpp
l-hoang: I like having it around to make it explicit that it's a class variable I'm referring to (though I guess in this case since it's a .cpp file for a class it's kind of implied since "graph" doesn't exist anywhere else).

libquery/src/AttributedGraph.cpp
l-hoang: Bad assertion that was checking if the user was constructing a sane query that ended up biting us when the a passed in query wasn't as expected (though correctness wise it doesn't matter as long as the user doesn't redefine a label on a node).

Will remove.

ddn0: Assertion-wise, if this is something on user-defined data, then it would be better to return an explicit error to the user because we can't guarantee what they will submit as a query. In general it would be good to be very defensive around user input.

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -24,23 +24,32 @@
...
- */
+
+//! Gets position of rightmost set bit from 0 to 31.
+//! @param n number to get rightmost set bit of
+//! @returns position of rightmost set bit, zero-indexed
 unsigned rightmostSetBitPos(uint32_t n);
 
 namespace galois::graphs {
+//! metadata with info on how to return matched nodes to the user
+struct QueryNodeReturnMetadata {
```
roshandathathri, line 38: Probably better declared with CompilerQueryNode as CompilerQueryReturnNode (metadata is not required).

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -24,23 +24,32 @@
...
+  //! ID of node in query graph to use
+  unsigned qid;
+  //! data type of property (if single property)
+  std::shared_ptr<arrow::DataType> d_type;
```
roshandathathri, line 42: Why not a vector of property names?

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -150,51 +149,67 @@ class AttributedGraph {
...
+  //! corresponds to that vertex (embedding order is matching order
+  //! @param node_data_to_return Metadata compiled from cypher query telling
+  //! the system what needs to be returned from the matches
+  //! @param num_matches Total number of matches in returned_embeddings;
+  //! used for now to return counts
+  void printNodeResults(
+      QueryEmbeddingQueueType& returned_embeddings,
+      const std::vector<VertexId>& query_vertex_to_matching_order,
+      const std::vector<QueryNodeReturnMetadata>& node_data_to_return,
+      size_t num_matches);
```
roshandathathri, line 193: This should NOT be a public function. It's better handled within `matchCypherQuery` which can take a boolean option whether to print the query results or not.

libquery/include/querying/CypherCompiler.h
```diff
@@ -67,10 +63,8 @@ struct CompilerQueryEdge {
   }
 };
 
-/**
- * Structure that holds the metadata for a return for a cypher call. Tells
- * program how to return something.
- */
+//! Structure that holds the metadata for a return for a cypher call. Tells
+//! program how to return something.
 struct CompilerQueryResult {
```
roshandathathri, line 68: Why not combine this and the other return structure?

libquery/src/AttributedGraph.cpp
```diff
@@ -583,11 +802,25 @@ AttributedGraph::matchQuery(std::vector<galois::CompilerQueryNode>& queryNodes,
...
-    return subgraphQuery(queryGraph, this->graph);
+    size_t num_matches =
+        subgraphQuery(queryGraph, this->graph, returned_embeddings,
+                      query_vertex_to_matching_order);
+
+    ////////////////////////////////////////////////////////////////////////////
+    // Return parsing
+    ////////////////////////////////////////////////////////////////////////////
+
+    printNodeResults(returned_embeddings, query_vertex_to_matching_order,
```
roshandathathri, line 818: Add a boolean flag to the matchQuery() function to enable or disable this. In the future, we might have an enum to allow different return formats. In the near future, it would be required to specify the output file/stream for printing.

libquery/src/SubgraphQuery.cpp
```diff
@@ -235,13 +232,13 @@ class SubgraphQueryMiner {
     } else {
       // this embedding has matched the entire query graph; increment
       // count
+      QueryEmbeddingType new_emb(emb);
+      new_emb.push_back(dst);
```
roshandathathri, line 236: These nested-ifs can be better organized. For example, move these two statements in both if and else before if.

libquery/src/SubgraphQuery.cpp
```diff
@@ -496,16 +496,21 @@ class SubgraphQueryMiner {
...
             addEmbedding<false, printEmbeddings>(0, emb, n, neighbors,
                                                  numInNeighbors, queue);
           }
         },
         galois::loopname("EmbeddingInit"));
 
-    // TODO handle 1 node query graph case here; everything in emb is a match
+    if (queryGraph.size() == 1) {
+      galois::gDebug("Single node query: returning");
+      returned_embeddings.swap(queue);
```
roshandathathri, line 511: This is fine. However, I would prefer to return the list of embeddings (instead of pass by reference argument).

libquery/src/SubgraphQuery.cpp
```diff
@@ -496,16 +496,21 @@ class SubgraphQueryMiner {
...
+      return;
+    }
 
     if (DFS) {
       extend_vertex<true, printEmbeddings>(queue, queue2);
```
roshandathathri, line 516: return_embeddings are not set in this path.

I would prefer a single return statement. This if-else block can be nested within the `if (queryGraph.size() > 1)` block and after that block, you set return_embeddings and return.

libquery/src/SubgraphQuery.cpp
```diff
@@ -530,15 +538,14 @@ class SubgraphQueryMiner {
 
   size_t get_total_count() { return total_count; }
 
-  void print_output() {
-    galois::gDebug("Number of matched subgraphs: ", get_total_count(), "\n");
+  std::vector<VertexId> getMatchOrder() {
```
roshandathathri, line 541: Why not return by reference?

Also, this is const function.

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: disagree; it's not needed at all in the compiler code, and can only be constructed after the query graph is constructed by an attributed graph function since it requires knowledge of which QID corresponds to which compiler id.

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: what do you mean by this?

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: makes sense; it's only called internally anyways. will move

libquery/include/querying/CypherCompiler.h
l-hoang: One requires knowledge of the constructed query graph *after* compilation.

libquery/src/SubgraphQuery.cpp
l-hoang: Returning the list as a regular QueueType would require the insert bag to be copyable. if i recall correctly it isn't (for good reason; you don't want to copy pointers to linked lists that insert bag is maintaining)

Returning the list as a *reference* to a query type would require that queue or queue2 be backed by persistent memory (since they're declared in this function, they're backed by stack memory; returning a reference to stack memory is asking for trouble).

Neither option is really desirable here in my opinion. if there's something I'm missing, let me know.

libquery/src/SubgraphQuery.cpp
l-hoang: I was ignoring the DFS path at the moment; a single return statement makes sense

libquery/src/SubgraphQuery.cpp
l-hoang: I was thinking that this was a good idea, but then I realized that the miner object goes out of scope after the subgraph query function, so returning by reference is actually a bad idea since the object is not maintained after subgraphQuery is returned from.

l-hoang:
Pushed a few changes to resolve some of Roshan's comments. Should be ready to merge pending another review from Roshan.

libquery/include/galois/graphs/AttributedGraph.h
roshandathathri: Why are you identifying a property by its datatype instead of name/id?

l-hoang: the property is defined in the compiler_info field in the same struct; the datatype is used for casting the arrow array

libquery/src/SubgraphQuery.cpp
roshandathathri: https://stackoverflow.com/questions/753312/returning-large-objects-in-functions

You can return large objects without a reference and it won't be copied.

l-hoang: That's assuming you have a move constructor defined. I can check if insert bag has that defined.

libquery/src/SubgraphQuery.cpp
roshandathathri: It can still be const, right?

libquery/src/SubgraphQuery.cpp
roshandathathri: This should work. If it doesn't, we should fix insert bag to make it work.

libquery/src/SubgraphQuery.cpp
l-hoang: yes; i made it a const function.

libquery/src/SubgraphQuery.cpp
l-hoang: confirmed that InsertBag has a move constructor and that it's being used. pushing in the updated version now

l-hoang:
Embeddings now returned via move semantics (thanks @roshandathathri ); should be good to go after one last check + CI passes.

libquery/src/AttributedGraph.cpp
```diff
@@ -364,129 +375,318 @@ size_t AttributedGraph::matchCypherQuery(const char* cypherQueryStr) {
...
+      new_metadata.compiler_info = cur_result;
+      if (!cur_result.is_count) {
+        if (cur_result.property_name.empty()) {
+          // not a count + no property name = need to return all props
+          new_metadata.all_props = true;
+        } else {
+          // not a count + has a property name; get the property type
+          new_metadata.all_props = false;
+          new_metadata.d_type =
+              graph.getNodeAttributeType(cur_result.property_name);
```
roshandathathri, line 434: Don't see why you need to maintain to separate structure for this information. All of this is derivable from CompilerQueryResult with a simple map (from name to index or from vector index to query node index). Why can't this be done on the fly without maintaining a structure?

l-hoang: it can be done on the fly

for the sake of modularity i designed it this way because doing it on the fly in the return part of the code would make that code harder to understand (IMO) than it needs to be, and I personally prefer abstracting a "return" value into a single struct with as little external dependency as possible

PR #200, opened 2020-07-12
tsuba: Re-orgainize low level code into `FileStorage` objects

### LINKED ISSUE #193: Remove `int tsuba::FileOpen(const std::string& uri);`
It is no longer a supported way to use tsuba. Right now it only exists because we call it in `FileGraph`.

tylershunt:
Tsuba has different `FileStorage` objects per back end and chooses the
correct one to satisfy a storage request by examining the URI. Curently:

s3://...  uris go to s3
file://... and all other uris go to the local file system

Seems like a worthwhile step forward to avoid ad hoc URI checks.

After #199

Closes #193 since I was cleaning up the interface anyway, it means that
FileGraph can no longer read from S3 but I don't think anyone cares.


libtsuba/src/S3Storage.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+galois::Result<void> S3Storage::Init() { return S3Init(); }
+
+galois::Result<void> S3Storage::Fini() { return S3Fini(); }
+
+galois::Result<void> S3Storage::Stat(const std::string& uri, StatBuf* s_buf) {
+  auto [bucket, object] = S3SplitUri(std::string(uri));
+  if (bucket.empty() || object.empty()) {
+    return ErrorCode::InvalidArgument;
+  }
```
witchel, line 18: I wonder if it is worth having a function similar to CleanUri that all S3 functions call in the start.  Maybe
auto [res, bucket, object] = S3CleanUri(uri);
if(!res) { return Res.error() }

libtsuba/src/file.cpp
```diff
@@ -11,23 +11,16 @@
 #include <fstream>
 #include <iostream>
 #include <cassert>
-#include <boost/filesystem.hpp>
```
witchel, line 14: Oh yeah!  Bye boost/filesystem.ccp from file.

libtsuba/src/S3Storage.cpp
tylershunt: that's a good idea.

libtsuba/include/tsuba/Errors.h
```diff
@@ -16,6 +16,7 @@ enum class ErrorCode {
   Exists          = 5,
   OutOfMemory     = 6,
   TODO            = 7,
+  StorageError    = 8,
```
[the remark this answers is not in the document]
tylershunt: Yes, i think all of the current `StorageError`s are s3 errors anyway. So that seems reasonable. Yeah all of the local file system errors are propagated by `ResultErrno`s.

PR #203, opened 2020-07-13
tsuba: Provide quick unit tests

witchel:
This provides a quick unit test for tsuba that covers file creation, copy and reading for local and S3 files.

I also split galois/Random.h from galois/FileSystem.h in libsupport.

I modify the PATH and AWS_DEFAULT_REGION in the environment in a way that might be not so good.  I'm assuming unit tests are run from the build directory, but that also might be flawed.


ddn0:
From CMake, you can set the environment variables, working directories, etc. That might be a more reasonable place for this type of configuration since it is the framework responsible for spawning the tests in the first place. Likewise, you can use the generator target file [1] to pass the full path of any auxiliary executables.


[1] https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#target-dependent-queries

witchel:
> From CMake, you can set the environment variables, working directories, etc. That might be a more reasonable place for this type of configuration since it is the framework responsible for spawning the tests in the first place. Likewise, you can use the generator target file [1] to pass the full path of any auxiliary executables.
> 
> [1] https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#target-dependent-queries

You're a naughty one.  Saucy Jack.

I spent over an hour trying to get cmake to reliably set multiple environment variables.  I'm not even sure what it considers a valid list.  Documentation is sparse, as are examples.  It's a world of pain.

ddn0:
FWIW, if you need to set multiple environment variables, here's an example https://github.com/KatanaGraph/katana/blob/master/lonestar/CMakeLists.txt#L166

PR #204, opened 2020-07-13
Change galois::NewPath to return galois::Result<std::string>

scober:
I have a few questions about these changes:

1) By returning a string instead of a boost::filesystem::path, we now have to handle some things on our own. One of those things is appending a file name to a directory path. The current implementation in this PR is to unconditionally add a '/' character between the two strings. This is not the same as the old boost behavior, but is it sufficient for our uses?

2) The other thing we now have to do ourselves is extract file names from path names. The current implementation in this PR is to return the substring following the last '/' character in the path name. This is not the same as the old boost behavior, but is it sufficient for our uses?

3) Is there a proper way to initiate an error from inside libsupport? The other functions I found that in directory just propagate errors (i.e. galois::ResultErrno()). Is there a better way to initiate an error than generating a system error? (Which is what I did).


libsupport/src/FileSystem.cpp
```diff
@@ -80,12 +80,21 @@ std::string generate_random_alphanumeric_string(std::size_t len) {
...
-boost::filesystem::path galois::NewPath(const boost::filesystem::path& dir,
-                                        const std::string& prefix) {
-  std::string name = prefix;
+galois::Result<std::string> galois::NewPath(const std::string& dir,
+                                            const std::string& prefix) {
+  std::string name = "/" + prefix;
   name += "-";
   name += generate_random_alphanumeric_string(12);
-  boost::filesystem::path p{dir};
+  std::string p{dir};
```
tylershunt, line 88: One of the nice things that boost was buying us was handling cases where `dir` ends with `/` and `prefix` starts with `/`. These will get ironed out by Linux but I think will result in some "interesting" file names in `s3`.  I think this function should handle at least those cases in  a sane way (or else there's not much value over managing the path's ad hoc elswhere).

libsupport/src/FileSystem.cpp
```diff
@@ -80,12 +80,21 @@ std::string generate_random_alphanumeric_string(std::size_t len) {
...
   return p.append(name);
 }
+
+// This function does not recognize any path seperator other than '/'. This
+// could be a problem for Windows or "non-standard S3" paths.
```
tylershunt, line 93: Fwiw I think this is fine. We're treating Unix style path separators as "directories" in S3.

libsupport/src/FileSystem.cpp
witchel: Agreed.  Bonus for defining kSepChar as '/' or kSepString as "/"

libsupport/src/FileSystem.cpp
```diff
@@ -80,12 +80,21 @@ std::string generate_random_alphanumeric_string(std::size_t len) {
...
+galois::Result<std::string> galois::ExtractFileName(const std::string& path) {
+  size_t last_slash = path.find_last_of('/', std::string::npos);
+  if (last_slash == std::string::npos) {
+    return std::error_code(EINVAL, std::system_category());
+  }
```
witchel, line 98: Oh, I think this is what Simon has asking about.  Tyler, we've been returning our own error codes, not reusing system error codes, by and large.  Should he do this or just return ErrorCode::InvalidArgument from tsuba/Error.h?

tylershunt: I think this is a fine stop-gap for now; the principled way to solve this is to define an `ErrorCode` type in libsupport and throw one of those errors. @ddn0 since he's the error master mind and can tell y'all if i'm off my rocker here.

tylershunt:
I totally just read code without looking at your questions 😨. But if it wasn't clear from my comments the answers are:
 1. No
 2. Yes
 3. Probably fine but not technically the right way to do it.
😛

libsupport/src/FileSystem.cpp
ddn0: I agree with @tylershunt ! There is a more principled way; though, it doesn't seem worth the machinery now.

Also, I believe you can just do: `return std::errc::invalid_argument`

libsupport/src/FileSystem.cpp
scober: So it is safe to say that '/' is the one true separator, but we also need to assume that S3 paths are canonically-formed?

libsupport/src/FileSystem.cpp
scober: Why do we define a different `ErrorCode` type in each lib?

libsupport/src/FileSystem.cpp
tylershunt: Yes

libsupport/src/FileSystem.cpp
tylershunt: Because different libs have different errors (even if there is overlap); and the type of `ErrorCode`  makes checks like `is tsuba error` or `is libsupport error` easier.

tylershunt: (And there is machinery for relating errors in to general types, i.e., the `default_error_condition` function defined in `libgalois/include/galois/ErrorCode.h` and `libtsuba/include/tsuba/Errors.h`

scober: I went ahead and just added the machinery for principled errors.

PR #212, opened 2020-07-14
ci: Correct some drift between macOS and ubuntu builds

### LINKED ISSUE #63: Investigate CI failure on MacOS 10.15
I can build on my Mojave machine (10.14) but CI dies on Catalina (10.15) with:
```
ld: framework not found -lCocoa
```

ddn0:
Closes #63


witchel:
This is probably not the right spot, but I am wondering about the Catalina build environment as I will be switching to laptop use for a few days.  I'm not sure if Issue #63 means we have MacOS build issues.  I don't see a setup_dev_macos.sh, in analogy with setup_dev_ubuntu.sh.

ddn0:
> This is probably not the right spot, but I am wondering about the Catalina build environment as I will be switching to laptop use for a few days. I'm not sure if Issue #63 means we have MacOS build issues. I don't see a setup_dev_macos.sh, in analogy with setup_dev_ubuntu.sh.

You read my mind!

I'm thinking of reorganizing this PR to re-enable the MacOS CI and bring all the setup inline with our Linux builds. I was able to build on a Catalina machine so I think #63 must have to do with how the GitHub runner is configured rather than Catalina itself.

ddn0:
@scober `mmap(..., MAP_32BIT)` is not available on MacOS. I just removed it, but let me know if I should attempt a better workaround.

scober:
> Open questions:
> 
> - @scober `mmap(..., MAP_32BIT)` is not available on MacOS. I just removed it, but let me know if I should attempt a better workaround.
> 

@patrickkenney9801  was going to remove this too as it was causing OOM errors in large graph conversions. It is not necessary for correctness, just an ill fated performance hack.

patrickkenney9801:
yes, removing MAP_32BIT is also in my current PR

tylershunt:
> * I'm tempted to enable the MacOS builds for now but disable the tests until we can reproduce this more easily.

That seems reasonable to me.

libtsuba/tests/md5.cpp
```diff
@@ -7,7 +7,9 @@
 
 #include "md5.h"
 
-#ifndef _MSC_VER
+#if defined(__APPLE__)
+#include <machine/endian.h>
+#elif !defined(_MSC_VER)
```
tylershunt, line 12: Just curious: Does the build work with `cl`? I've never tried it.

ddn0: @insertinterestingnamehere has probably tried most recently. I believe the current status is the build doesn't work on Windows.

arthurp:
I can follow this PR with a PR to enable MacOS conda packaging. Once builds on MacOS are working Conda packaging should be straight forward.

PR #219, opened 2020-07-15
gr2neo4j: update + output less unnecessary fluff

l-hoang:
This PR includes some minor changes I made to the existing gr2neo4j converter 
to get it working for the July 15 deadline.

Cleans up gr2neo4j code to work with neo4j based on new knowledge on how
import into neo4j works.

Also, previously gr2neo4j was outputting a lot of unnecessary detail that can
be specified during neo4j import time rather than contained in the file
(which makes the file bigger). This has been removed. I also removed the
dataAsLabel option because having a label for every single edge weight
is quite inefficient (it was being used for a 3rd party system if I
recall correctly that Katana will not need).


tools/graph-convert/graph-convert.cpp
```diff
@@ -2691,12 +2686,6 @@ struct Gr2Neo4j : public Conversion {
     using EdgeData        = galois::LargeArray<EdgeTy>;
```
gurbinder533, line 2686: It will be really useful if you can write a brief description about the different files that are produced by this conversion and maybe a sample command line to use those with neo4j.

l-hoang:
Updated with documentation; will merge once CI passes unless there are objections

PR #227, opened 2020-07-16
ci: Reduce memory usage by reducing parallelism

ddn0:


ddn0:
Let me check this by forcing a clean build. The CI checks don't really make sense here because it will pull from the cache and not build at all.

ddn0:
Example of a successful run from cold cache: https://github.com/KatanaGraph/katana/pull/227/checks?check_run_id=882721343

PR #228, opened 2020-07-16
Converter: Neo4j import sparse data

patrickkenney9801:
This PR optimizes the converters memory performance by providing some null compression for arrow arrays
This PR also parallelizes some components of the converter
This PR is also intended to clean up the code base to make it easier to use and understand, suggestions needed


l-hoang:
What's the difference between this PR and the earlier one you posted?

patrickkenney9801:
It provides significant memory optimizations that allow conversions of extremely sparse data that was otherwise impossible due to arrow's representation of it, i.e. nbdc dataset that has 39K labels.
It adds some parallelization
It has significant refactoring and I would like more advice on how to refactor further to make the code more usable by others

l-hoang:
I guess a better way to phrase my question is this:

Is the older PR a subset of this PR, or do they contain different things?

patrickkenney9801:
The old PR has been merged in master and the history has been rewritten properly for this PR now Loc.

ddn0:
Can you add some performance numbers before and after?

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -395,31 +457,12 @@ std::shared_ptr<arrow::Array> buildArray(std::shared_ptr<T> columnBuilder) {
...
 ChunkedArrays buildChunks(std::vector<ArrowArrays>* arrayChunks) {
   ChunkedArrays chunkedArrays;
-  chunkedArrays.reserve(arrayChunks->size());
-  for (size_t i = 0; i < arrayChunks->size(); i++) {
-    chunkedArrays.push_back(
-        std::make_shared<arrow::ChunkedArray>(arrayChunks->at(i)));
+  chunkedArrays.resize(arrayChunks->size());
+  for (size_t n = 0; n < arrayChunks->size(); n++) {
+    chunkedArrays[n] =
+        std::make_shared<arrow::ChunkedArray>(arrayChunks->at(n));
```
nicelhc13, line 465: This loop looks parallelizable if memory allocating inside of the parallel loop would not be an issue.

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -1470,17 +1799,17 @@ bool processNode(xmlTextReaderPtr reader,
...
               // if an entry for the key does not already exist, make a default
               // entry for it
-              if (keyIter == nodeKeys->end()) {
-                addColumnBuilder(property.first, true, nodeKeys,
-                                 nodeColumnBuilders, nodeSchemaVector,
-                                 nodeChunks, arrow::utf8(), nodes, chunkSize);
-                keyIter = nodeKeys->find(property.first);
+              if (keyIter == builder->nodeProperties.keys.end()) {
+                addStringBuilder(property.first, &builder->nodeProperties);
+                keyIter = builder->nodeProperties.keys.find(property.first);
```
nicelhc13, line 1807: It is just minor and remind. Instead of find() again, you could return or get (keyIter->second) value from reference after you append the value.

tools/graph-convert/tests/graph-properties-convert-test.cpp
```diff
@@ -1081,13 +1081,15 @@ void verifyChunksSet(galois::GraphComponents graph) {
...
+  galois::SharedMemSys sys;
   llvm::cl::ParseCommandLineOptions(argc, argv);
 
   galois::GraphComponents graph{nullptr, nullptr, nullptr, nullptr, nullptr};
   switch (fileType) {
   case galois::SourceType::GRAPHML:
     graph = galois::convertGraphML(inputFilename, chunkSize);
-    break;
+    // break;
+    return 0;
```
nicelhc13, line 1092: I am just curious why it immediately returns. Is this because GraphML does not have test?

tools/graph-convert/graph-properties-convert.cpp
patrickkenney9801: I did parallelize this at one point but creating chunked arrays is a very fast operation so it seemed to me like parallelization was overkill here

nicelhc13:
Galois is using camel case style. Is there any specific reason to replace with snake case style?

l-hoang:
> Galois is using camel case style. Is there any specific reason to replace with snake case style?

We use snake case for variable names and camel case for function names.

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -385,650 +431,719 @@ extractHeaderCSV(csv::CSVReader* reader, ArrowFields* nodeSchemaVector,
...
 // special case for building boolean builders where the empty value is false,
 // not null
-void addFalseColumnBuilder(const std::string& column,
-                           std::unordered_map<std::string, size_t>* map,
-                           BooleanBuilders* columnBuilders,
-                           ArrowFields* schemaVector,
-                           std::vector<ArrowArrays>* arrayChunks,
-                           std::shared_ptr<arrow::DataType> type, size_t offset,
-                           size_t chunkSize) {
+void AddFalseColumnBuilder(const std::string& column, LabelsState* labels) {
```
nicelhc13, line 481: I can see mix-usage of pointers and references. Lets use references if we guarantee that we do not pass NULL is preferable in C++. No big preferences but just for readability and for safety.

tools/graph-convert/graph-properties-convert-main.cpp
```diff
@@ -8,39 +8,39 @@
...
                                     "source data came from mongodb")),
-             cll::init(galois::SourceDatabase::NONE));
+             cll::init(galois::SourceDatabase::kNone));
 static cll::opt<size_t>
-    chunkSize("chunkSize",
-              cll::desc("Chunk size for in memory arrow representation"),
-              cll::init(25000));
+    chunk_size("chunk_size",
+               cll::desc("Chunk size for in memory arrow representation"),
+               cll::init(25000));
```
nicelhc13, line 37: Chunk size should not be 0. It is actually dangerous to invoke divide by zero or other consecutive errors.
Please put condition checking this in the very beginning of the main or parsing part.

nicelhc13:
> > Galois is using camel case style. Is there any specific reason to replace with snake case style?
> 
> We use snake case for variable names and camel case for function names.

nicelhc13:
> Can you add some performance numbers before and after?

That's a very good idea. Patrick is doing sorting optimization. We will share a report including that data soon.

tools/graph-convert/graph-properties-convert.cpp
ddn0: I'll confess that I haven't been good about using `T&` for in/out parameters. It looks like that is the preferred way (after just using a return value):
  
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#fcall-parameter-passing

tools/graph-convert/graph-properties-convert-main.cpp
ddn0: also, for the name of the command line parameter, either stick with camelCase or go with kabeb-case both of which are more common than snake_case for command line parameters.

tools/graph-convert/graph-properties-convert.cpp
patrickkenney9801: @tylershunt was a fan of using pointers for mutated data, should I still change all the pointers I can to references?

tylershunt: I actually thought that pointers were the preferred way according to the style guide 😅, but I'm willing to say that I do prefer pointers for arguments that mutate data (and I think if you run `clang-tidy` with it will ask you do that, or I've miss-configured it).

ddn0: It is weird.

The check in question is: https://clang.llvm.org/extra/clang-tidy/checks/google-runtime-references.html though the link to the section in the Google style guide is dead. The only statement I can find about reference parameters is https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs which basically matches the C++ core guidelines.

I also like an explicit pointer for in/out parameters.

Since an explicit pointer is the current state of the code and that's what the clang-tidy configuration enforces, let's stick with that for now. If we want to change it, we should start by updating the clang-tidy configuration.

tylershunt: Ah I'm not insane: https://web.archive.org/web/20200101070340/https://google.github.io/styleguide/cppguide.html this was the rule at the beginning of the year.

ddn0: Here's the change: https://github.com/google/styleguide/commit/7a7a2f510efe7d7fc5ea8fbed549ddb31fac8f3e#diff-26120df7bca3279afbf749017c778545

ddn0:
Is this PR ready for review?

patrickkenney9801:
This PR is now ready for review

tools/graph-convert/graph-properties-convert.h
```diff
@@ -8,32 +8,32 @@
 
 namespace galois {
 
-enum SourceType { GRAPHML, JSON, CSV };
-enum SourceDatabase { NONE, NEO4J, MONGODB };
-enum ImportDataType { STRING, INT64, INT32, DOUBLE, FLOAT, BOOLEAN };
+enum SourceType { kGraphml, kJson, kCsv };
+enum SourceDatabase { kNone, kNeo4j, kMongodb };
+enum ImportDataType { kString, kInt64, kInt32, kDouble, kFloat, kBoolean };
```
l-hoang, line 13: my guess is that the style guide recommends enums in this case. if that's the case, ignore this comment; I just find this casing for enums strange (used to all caps)

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -27,79 +28,127 @@
...
+  size_t falses_added;
 };
 
 /************************************/
 /* Shared GraphML/CSV/JSON Function */
 /************************************/
 
 template <typename T>
-void computePrefixSum(std::vector<T>* array) {
+void ComputePrefixSum(std::vector<T>* array) {
```
l-hoang, line 121: Galois provides a parallel prefix sum function in its parallel stl library; depending on how large this vector is it may or may not matter.

tools/graph-convert/graph-properties-convert.h
patrickkenney9801: yes

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -238,10 +284,10 @@ void addFalseColumnBuilder(const std::string& column,
 template <typename T>
 void addColumnBuilder(const std::string& column,
```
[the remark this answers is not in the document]
patrickkenney9801: I think according to @ddn0 we are using UpperCamelCase for function names unless they are getters or mirror std library functions but I am not sure. The functions that are camelCase should all be csv/json functions which are not used, but I will fix them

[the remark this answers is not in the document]
patrickkenney9801: I added some comments for graphml functions if you want to take a look and maybe give some pointers Loc

patrickkenney9801:
Here is a link to some preliminary data on this PR
https://docs.google.com/spreadsheets/d/1CC2NKKXO1OZXnC5Pp6dryXdwodNy-HZoNBP5aU37kNQ/edit#gid=0

PR #229, opened 2020-07-17
Add function to partially load a parquet file

scober:
This implementation does not reduce memory use or time spent interacting with storage (it still loads the entire file into memory). It does reduce computation and time spent to convert parquet into arrow which (we have learned) is a non-trivial portion of the total cost of this operation.
The "old way" of performing this operation is to convert a whole parquet file and slice out the desired table. This PR converts only the necessary portions of the parquet file (at row group granularity).
I ran a simple benchmark where I read a 500 MB arrow table out of a 4GB parquet file both ways:

| version | total time | conversion time | s3 time |
|---------|------------|-----------------|---------|
| old     |       6.98 |            3.67 |    3.32 |
| PR      |       4.30 |            0.92 |    3.38 |

I also wrote a unit test that just ensures correctness, but I had to export my `LoadPartialFile` function in RDG.h to make that work. If anyone knows of a way to test the function without exposing it, I would love to do that. A PR to reduce memory use and storage time will follow.


libtsuba/src/RDG.cpp
```diff
@@ -871,3 +871,62 @@ galois::Result<void> tsuba::DropEdgeProperty(RDG* rdg, int i) {
...
+
+// This shouldn't actually be used outside of this file, just exported for
+// testing
+galois::Result<std::shared_ptr<arrow::Table>>
+tsuba::LoadPartialTable(const std::string& expected_name,
+                        const std::string& file_path, int64_t offset,
+                        int64_t length) {
+
+  if (offset < 0 || length < 0) {
+    return tsuba::ErrorCode::InvalidArgument;
```
witchel, line 883: Why is offset a signed quantity?  With system APIs (like mmap), it usually is unsigned.
I mean, there isn't a good logical reason for length to be signed, but npos was -1, at least in one implementation I saw.  Maybe that is an arrow thing?

libtsuba/tests/fo_bench.cpp
```diff
@@ -325,11 +451,43 @@ void ReadWriteFO(std::shared_ptr<arrow::Table> table, std::string bucket,
...
+  int64_t offset = dist(gen);
+
+  std::shared_ptr<arrow::Table> recovered;
+  switch (version) {
+  case 0:
+    recovered = ReadPartial_v0(path, offset, length);
+    break;
+  case 1:
+    recovered = ReadPartial_v1(path, offset, length);
+    break;
```
witchel, line 477: It is nice to have the code to build the partial table in both ways.

libtsuba/src/RDG.cpp
scober: offset and length are signed to match the corresponding arguments in arrow::Table::Slice since this is supposed to replace that call in some sense. I don't know why they are signed in arrow, I think their typing choices are strange in a lot of places.

libtsuba/tests/fo_bench.cpp
scober: My thought was this benchmark could hold historical versions of this logic so we have a baseline.

scober:
@witchel Do you have any thoughts on unit testing `LoadPartialTable` without exposing it? It would be a helper function internal to RDG.cpp if I had some other way of testing it.

witchel:
> @witchel Do you have any thoughts on unit testing `LoadPartialTable` without exposing it? It would be a helper function internal to RDG.cpp if I had some other way of testing it.

I put some S3 functions in the tsuba::internal namespace.  I'm not sure if that will work for you, but I think it sends a clear message.

scober:
I’ll check that out, that sounds like a good option.

Issue #230, opened 2020-07-17
Sort arrow property arrays when sorting edges


l-hoang:
The Cypher graph data structure sorts edges by destination/label for efficiency. The arrow property arrays for the edges are not sorted along with them at the moment, leading to a mismatch between an edge and its property. It causes things like this:

```
||  (:CONTAINER_OF  {} )  ||
||  (:CONTAINER_OF  {joinDate: 2017-08-28T06:24:31.056Z,  } ) ||
||  (:CONTAINER_OF  {} )  ||
||  (:CONTAINER_OF  {} )  ||
||  (:CONTAINER_OF  {joinDate: 2013-06-07T01:11:52.767Z,  } ) ||
||  (:CONTAINER_OF  {joinDate: 2014-12-12T20:13:48.712Z,  } ) ||
||  (:CONTAINER_OF  {} )  ||
```

In LDBC data, CONTAINER_OF edge types should not have any attribute attached with them, but in this case, they do.

This can be resolved by either sorting the arrow property arrays along with the edge destinations or by mapping the resorted edge index to the original index in order to index into the array correctly. I am currently working on the former, but if there are objections, this is the place to discuss them.

PR #231, opened 2020-07-17
Return edge labels/attributes via Cypher

l-hoang:
Users can now write return <x> where x is an edge variable, and the
engine will print the data as requested. This PR also contains
miscellaneous additions to support this feature, including a function
to get a string representing the labels of an edge, changes to the
LabeledGraph's binary search function to return the found edge iterator,
and a findEdge function in LabeledGraph that, unlike the existing
isConnected function, will return the found edge iterator.

Note that return attributes printed are currently incorrect due to an
issue found while adding this feature in; see issue #230 for context. A
PR to fix this is coming soon.


libgalois/include/galois/graphs/LC_CSR_Labeled_Graph.h
```diff
@@ -475,15 +475,23 @@ class LC_CSR_Labeled_Graph
...
+      else
+        return false;
     } else {
-      auto begin = edge_begin(search, data);
-      auto end   = edge_end(search, data);
-      return binarySearch<false>(key, begin, end);
+      auto begin                     = edge_begin(search, data);
+      auto end                       = edge_end(search, data);
+      std::optional<edge_iterator> r = binarySearch<false>(key, begin, end);
+      if (r)
```
roshandathathri, line 491: Move the common if-else return block to after the search if-else block.

libgalois/include/galois/graphs/LC_CSR_Labeled_Graph.h
```diff
@@ -506,6 +514,33 @@ class LC_CSR_Labeled_Graph
     return false;
   }
 
+  /**
+   * Returns an edge iterator to an edge with some source and destination.
+   * If not found, returns nothing.
+   */
+  const std::optional<edge_iterator> findEdge(GraphNode src, GraphNode dst) {
```
roshandathathri, line 521: It would be good to specify in the documentation above that you are searching for dst in src's edges.

libquery/src/AttributedGraph.cpp
```diff
@@ -453,76 +467,134 @@ void AttributedGraph::printNodeResults(
...
+          uint32_t dst_qid = get_src_dst_pair_result->second.second;
+          // matching order for those qids
+          uint32_t src_match_order = query_vertex_to_matching_order[src_qid];
+          uint32_t dst_match_order = query_vertex_to_matching_order[dst_qid];
+          // from matching order, can get actual ID
+          uint32_t src_data_id = emb.get_element(src_match_order).get_vid();
+          uint32_t dst_data_id = emb.get_element(dst_match_order).get_vid();
+
+          // finds the outgoing edge id
+          auto found_edge = this->graph.findEdge(src_data_id, dst_data_id);
```
roshandathathri, line 550: Don't you know the label of the edge to find? I think you need an if-else block to handle multiplexed query labels. If (multiplexed) label does not exist, then call this function; otherwise, call the find edge specific to the label.

l-hoang: That information currently isn't being saved by the compiler end into the ReturnValue structs, though it could be. Not for this PR though.

roshandathathri: Please add a TODO in code then. No need for a github issue.

l-hoang:
Updated

PR #236, opened 2020-07-17
Cypher: array access account for sorted edges

### LINKED ISSUE #230: Sort arrow property arrays when sorting edges

l-hoang:
Before this commit, arrow array accesses for edges were incorrect as the
underlying CSR graph is sorted for efficiency while the arrow arrays are
not sorted along with them. This causes references to a particular edge
index to be incorrect. This commit adds a mapping from original edge
index to new edge index that's used when accessing the arrow arrays; it
is constructed by sorting a 1-to-1 mapping in the same manner as the CSR
is sorted.

The original plan was to sort all of the arrow arrays to avoid a layer
of indirection for each edge property access, but complications there
included creating a new arrow chunked array for the sorted array which
included having to deal with null elements during the sort. This is
not to say this isn't doable; in fact, the original to new edge mapping
created in this commit can be used to facilitate the creation of new
arrow arrays for this purpose. That can be done in another PR.

Resolves #230


libgalois/include/galois/graphs/LC_CSR_Labeled_Graph.h
```diff
@@ -511,6 +511,48 @@ class LC_CSR_Labeled_Graph
...
 
+  /**
+   * Given some vector, sort the indices of that vector as if they were
+   * the edge destinations that would get sorted if one sorted by the
+   * edge data then destinations. Used mainly to rearrange other vectors
+   * that also need to be sorted besides destinations/data  (since the
+   * current infrastructure only supports sorting those 2 arrays at
+   * the moment).
+   */
+  void sortVectorByDataThenDst(std::vector<uint64_t>& vector_to_sort) {
```
roshandathathri, line 522: I think this should be a private function and the vector should be a member variable. We should have a function which takes an iterator and returns an iterator. This function should map the edge from old index to new index.

roshandathathri: Also, we would two such private functions and two vector member variables - one for outgoing edges and one for incoming edges.

libquery/include/querying/IndexableChunkedArray.h
```diff
@@ -66,6 +76,11 @@ class IndexableChunkedArray {
   getNumericElement(int64_t global_index) const {
     // user is responsible for correct numeric type outside of debug build
     assert(this->getType()->Equals(NumType()));
+    assert((int64_t)this->index_to_sorted_edge.size() == this->size());
+
+    // get the remapped sorted index for this edge
+    global_index = this->index_to_sorted_edge[global_index];
```
roshandathathri, line 82: This should be done outside the arrow interface.

libquery/src/IndexableChunkedArray.cpp
```diff
@@ -68,6 +70,11 @@ std::optional<std::string>
 IndexableChunkedArray::getStringElement(int64_t global_index) const {
   // user is responsible for calling appropriate function outside debug
   assert(this->getType()->Equals(arrow::utf8()));
+  assert((int64_t)this->index_to_sorted_edge.size() == this->size());
+
+  // get the remapped sorted index for this edge
+  global_index = this->index_to_sorted_edge[global_index];
```
roshandathathri, line 76: Ditto.

l-hoang:
re: outgoing/incoming; I haven't thought about how to do incoming, but what you said makes sense. I haven't thought about this much, but I think you can get away with just one remapping array since the incoming -> outgoing might already exist from LC_CSR_CSC, no? (unless we aren't using that feature)

Also agreed about moving the reindexing out of the arrow array, especially since as you mentioned I'm not considering incoming edges yet and that would screw up how that works if it's kept as is.

In any case, isn't the idea to abandon CSR_Labeled graph soon anyways for PropertyGraph? (as in I hope the next week or so :P)

PR #250, opened 2020-07-21
Add support for uint8_t edge properties to CuSP. 

gurbinder533:


libcusp/include/galois/graphs/NewGeneric.h
```diff
@@ -492,6 +491,27 @@ class NewDistGraphGeneric : public DistGraph<NodeTy, EdgeDataTy> {
         columns.push_back(array);
         break;
       }
+      case arrow::Type::UINT8: {
+        auto array_builder = std::make_shared<arrow::UInt8Builder>();
```
roshandathathri, line 495: I suggest making the cases templatized now rather than later.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
```diff
@@ -67,12 +67,17 @@ const uint32_t infinity = std::numeric_limits<uint32_t>::max() / 4;
 enum {
   NODE_DATA_DIST_CURRENT,
 };
+enum {
+  EDGE_DATA_EDGE_WEIGHT,
+};
 
 using NodeData = std::tuple<uint32_t>;
+using EdgeData = std::tuple<int64_t>;
```
roshandathathri, line 75: Why uint64_t when distances are uint32_t?

lonestar/analytics/distributed/sssp/sssp_pull.cpp
```diff
@@ -67,12 +67,17 @@ const uint32_t infinity = std::numeric_limits<uint32_t>::max() / 4;
 enum {
   NODE_DATA_DIST_CURRENT,
 };
+enum {
+  EDGE_DATA_EDGE_WEIGHT,
```
roshandathathri, line 71: Is the input graph expected to have this field?

lonestar/analytics/distributed/sssp/sssp_pull.cpp
l-hoang: probably because if you export from neo4j there's no concept of an unsigned int, so the parquet type is a int64.

You could probably cast it while reading, but I'm not sure what happens then + that sounds quite brittle/dangerous.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
l-hoang: not necessarily with this name; it just needs something with type int64. This name is user defined and can be whatever you want it to be.

roshandathathri: Then where is the name for the property specified?

libcusp/include/galois/graphs/NewGeneric.h
l-hoang: I agree with the reason why you would want this, but I don't think it's possible.

The field type isn't known at compile time, so the compiler would complain about not knowing the type.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
l-hoang: command line option at runtime if I recall correctly

libcusp/include/galois/graphs/NewGeneric.h
roshandathathri: No, I'm NOT asking for the switch cases to be removed. The cases will remain. The code within each case can be templatized right?

l-hoang: with multiple template arguments, yes, which would be better than the code duplication going on now

lonestar/analytics/distributed/sssp/sssp_pull.cpp
gurbinder533: Yes, the "value" edge property on converted rmat15 is int64 actually.

libcusp/include/galois/graphs/NewGeneric.h
gurbinder533: Good idea. Will do.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
gurbinder533: This enum is just to get the index of the appropriate edge property from the EdgeData tuple. Currently, the edge property name is specified at the command line (only one property at the moment). I did this in order to load multiple edge-properties but I am still not sure how to connect the enum names used for index with the property names specified on command line.

roshandathathri: Okay. The edge property name was part of previous PR.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
roshandathathri: This will hurt application performance. Can we not convert it uint32_t while loading? The computation (addition) will be done in int64_t and then type-casted to uint32_t, which is an overhead.

libcusp/include/galois/graphs/NewGeneric.h
gurbinder533: Done.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
l-hoang: yes, you can by creating another arrow array of type uint32t and writing out the int64 array to it. Not sure how this would hurt write support down the line (e.g. if you have a 32t in memory and edit it, you'll have to write it out later as an int64)

I propse this be handled in another PR.

roshandathathri: That's fine with me.

gurbinder533:
I did minimal changes to compile GPU path for now but there is a bigger question of how to support multiple edge properties for GPUs.

gurbinder533:
In lingpu/include/csr_graph.h, the typedef for node_data_type and edge_data_type in line 23 ans 24 ([link](https://github.com/KatanaGraph/katana/blob/0cc3bfb05777b62edfdaf7bb56dcd6ef53c68ab7/libgpu/include/csr_graph.h#L23)), anything other than "int" gives me the following weird error:

lonestar/analytics/gpu/bfs/bfs.cu(14): error: invalid **redeclaration** of type name "edge_data_type"
libgpu/include/csr_graph.h(23): here

lonestar/analytics/gpu/bfs/bfs.cu(15): error: invalid **redeclaration** of type name "node_data_type"
libgpu/include/csr_graph.h(24): here

2 errors detected in the compilation of "/tmp/tmpxft_000026f4_00000000-6_bfs.cpp1.ii".

Do we know what is happening here and was that the reason they were _int_ in the first place?

roshandathathri:
> lonestar/analytics/gpu/bfs/bfs.cu(14): error: invalid **redeclaration** of type name "edge_data_type"
> libgpu/include/csr_graph.h(23): here
> 
> lonestar/analytics/gpu/bfs/bfs.cu(15): error: invalid **redeclaration** of type name "node_data_type"
> libgpu/include/csr_graph.h(24): here

You can remove those redeclarations in bfs.cu pagerank.cu and sssp.cu.

lonestar/analytics/distributed/sssp/sssp_pull.cpp
ddn0: This sounds more like a limitation of whatever is converting the rmat15 file. It is certainly possible to read and write 32-bit integers even now. But if the schema says the edge data is truly 64 bit, it wouldn't be great to truncate it without some option to opt out.

gurbinder533: I agree with @ddn0. The edge property types should be dictated by the schema.

roshandathathri: Then we should use uint32_t for edge weights and only allow properties that have that type to be treated as edge weights.

> probably because if you export from neo4j there's no concept of an unsigned int, so the parquet type is a int64.

We should fix this to make sure that rmat15 has uint32_t edge weights.

I'm fine if this is done in a subsequent PR.

PR #266, opened 2020-07-25
Remove deprecated iss/galois scripts

gurbinder533:
This is to clean the scripts folder. All the old iss/galois scripts have been removed as we won't need them going forward. 
Only scripts that we can still use for katana are left. 

- Everything in the experimental folder has been deleted except for Pangolin verify scripts, which according to me can also be deleted from katana repo. Correct me if I am wrong.

- Few old vtune scripts are left.

Let me know if I have deleted more than I should or if we can delete more.


ddn0:
@insertinterestingnamehere if no one is using `make_dist.sh` for releases, we can also remove `make_dist.sh.in` and the related CMake machinery (scripts/CMakeLists.txt).

roshandathathri:
I think we are preemptively removing some scripts which may turnout be useful. "iss_load_modules" is one such example. I created most of the heterogeneousGalois scripts and I use them, especially to run on Stampede/Bridges. Some of the log parser scripts also fall in this category. These are "Galois" scripts, not "Katana" scripts but it is easy to port them. 

If we intend to separate out end-user and developer scripts, we should have separate sub-folders for them, instead of deleting all developer scripts.

I'm definitely not in favor deleting vtune, plot, and log-parser scripts.

ddn0:
I think the current set of removals is conservative, so let's just concentrate on removing these now.

I understand that there might be some outstanding uses, but I don't think the negative impact is that high considering that ISS/Galois still exists. In general, having untested, out-of-date or unmaintained code is a tarpit for developers. The best way to keep something in the codebase is to have a plan for its continual maintenance.

gurbinder533:
@roshandathathri I have tried to remove scripts that I don't see being used for katana in near future (for example job scheduling scripts on Stampede/Bridges) and keeping them around without maintaining is not good. And as @ddn0 said we have a backup for these in ISS/Galois anyways.

roshandathathri:
> I don't think the negative impact is that high considering that ISS/Galois still exists

Yes, so, it's fine. 

> I don't see being used for katana in near future

I agree that this is a good rubric.

gurbinder533:
I think we have a consensus then. I will merge this in some time. Let me know if that is not the case.

l-hoang:
For the record, I'm on the side of "remove everything that doesn't have to do with Katana" as mentioned earlier, but merging this in is fine for now.

PR #268, opened 2020-07-26
Support limited OrderedByIntegerMetric worklists in for_each

arthurp:
This PR also refactors the loop function dispatch to allow for easier modification. I attempted to document the new system in a comment in _loop.pyx.jinja.


python/galois/_loops.pyx.jinja
```diff
@@ -71,6 +75,34 @@ cdef class UserContext:
 {% endcall %}
 
 
+# FIXME: Hard coded uint64_t loop variable and int64_t order variable
```
gurbinder533, line 78: hardcoded uint64_t here is representing the node id, right?

python/galois/sssp_property_graph.py
```diff
@@ -0,0 +1,112 @@
...
+            ctx.push(dst)
+
+
+@obim_metric()
+def obim_indexer(shift, dists, nid):
+    return dists[nid] >> shift
+
+
+def sssp(graph: PropertyGraph, source, length_property, shift, property_name):
+    dists = create_dist_array(graph, source)
```
gurbinder533, line 39: create_dist_array --> create_distance_array?

python/galois/sssp_property_graph.py
```diff
@@ -0,0 +1,112 @@
...
+    parser.add_argument('--propertyName', type=str, default="NewProperty")
+    parser.add_argument('--lengthProperty', type=str, required=True)
+    parser.add_argument('--shift', type=int, default=6)
+    parser.add_argument('--reportNode', type=int, default=1)
+    parser.add_argument('--noverify', action='store_true', default=False)
+    parser.add_argument('--threads', '-t', type=int, default=1)
+    parser.add_argument('input', type=str)
+    args = parser.parse_args()
+
+    from galois.shmem import *
```
gurbinder533, line 100: We don't need this import anymore.

python/galois/sssp_property_graph.py
```diff
@@ -0,0 +1,112 @@
...
+    print("Max distance:", max_dist.reduce())
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--startNode', type=int, default=0)
+    parser.add_argument('--propertyName', type=str, default="NewProperty")
+    parser.add_argument('--lengthProperty', type=str, required=True)
```
gurbinder533, line 92: Maybe rename this to "edgeProperty" or "edgeWeightProperty"?

python/galois/_loops.pyx.jinja
arthurp: Yes

gurbinder533: In the current shared-memory Galois, node ids are actually uint32_t (also known as v1 graph version). I don't expect this to cause any problems as such but we should keep this in mind for consistency.

PR #277, opened 2020-07-28
Order by support + refactoring of compiler structures

l-hoang:
1) adds a struct called QueryProperty that captures a very common func(varname.propname) construction found in Cypher queries. This struct also makes CompilerQueryResult obsolete as it also contains an alias field to represent return values as well.

2) Adds compile support for Order By: parses it all into a struct. The idea is that one struct will be created per "return" clause (RETURN and WITH).

3) Also contains some formatting changes (PascalCase instead of camelCase) to match Katana style guidelines.


l-hoang:
@roshandathathri Instead of making QueryResult store OrderBy and Metadata I dropped the class completely and just put return results in a vector with the metadata separate from it. For return metadata it makes sense in my opinion because there is always only a single return value, so a single class var is fine.

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -203,7 +203,7 @@ class AttributedGraph {
   void printResults(
       QueryEmbeddingQueueType& returned_embeddings,
       const std::vector<VertexId>& query_vertex_to_matching_order,
-      const galois::gstl::Vector<CompilerQueryResult>& return_values,
+      const galois::gstl::Vector<QueryProperty>& return_values,
```
roshandathathri, line 206: Use std::vector instead - this can be done later.

l-hoang: Used gstl here because memory allocation can occur dynamically, though now that I think about it you know the number of return values from the number of projections present on that AST node so you can do a reserve.

roshandathathri: Yes. This is not (allocated in) multi-threaded code and the vector size is not expected to be large, so std::vector is preferable.

l-hoang:
> I'm still in favor of dropping OrderByMetadata class and including its fields in ReturnMetadata. OrderByMetadata is just a pair of QueryProperty and bool, which can be a typedef in ReturnMetadata.

I'm fine with this. What I wasn't really fine with was overloading a single struct with everything (that was going to be CompilerResult which would have (1) a vector of results, (2) skip/limit,distinct, and (3) order by, all of which would be in different fields). Since CompilerResult has been removed, the results are in their own class var vector and Metadata remains its own class, I guess I don't mind adding 2 fields to it too much.

But I'll do that later because that still requires a painful change in the compiler ctests.

roshandathathri:
> I'll do that later because that still requires a painful change in the compiler ctests.

Yes, it's not critical.

PR #282, opened 2020-07-29
Extract partitioning arguments

ddn0:
In preparation for supporting multiple node and edge properties,
refactor command line processing to use a separate DistOptions object
that encapsulates all the options for constructing a distributed graph.


lonestar/libdistbench/include/DistBench/Input.h
```diff
@@ -118,289 +99,161 @@ extern cll::opt<std::string> edgePropertyName;
  * Graph-loading functions
  ******************************************************************************/
 
-template <typename NodeData, typename EdgeData>
-using DistGraphPtr =
-    std::unique_ptr<galois::graphs::DistGraph<NodeData, EdgeData>>;
+enum class GraphType {
+  CSR,     //! Compressed sparse row graph format, i.e. outgoing edges
+  CSC,     //! Compressed sparse column graph format, i.e. incoming edges
+  CSR_CSC, //! A symetric graph, i.e., CSR + CSC
```
roshandathathri, line 105: If you intend these types to capture Outgoing, Incoming, Symmetric edges, then I recommend using those names instead of CSR, CSC, CSR_CSC.

LC_CSR_CSC_Graph does not mean the graph is symmetric. CSR_CSC in that case just means that one can iterate over both outgoing and incoming edges.

gurbinder533:
LGTM.

ddn0:
> Question: is there really any performance different for having these things be template arguments? I find that it just makes the code extremely messy if you start using enable_if to do things.

For `IterateOutEdges`, it doesn't really matter because all it does is affect the selection of the partition policy (which is a template parameter).

As to whether partition policy needs to a template parameter, you probably could reduce the dependence to just certain member functions rather requiring the entire type to be dependent on the policy. One benefit of functions as template parameters is avoiding the cost of indirection (e.g., function pointer, virtual call), but depending on the API, that benefit might be minimal (e.g., bulk operation vs single operation). 

I haven't looked too much on how the partitioners work, so I can't say if that is better though.

PR #286, opened 2020-07-29
Cleanup forward declarations

ddn0:
Forward declarations are annoying. Let's minimize the need for them if possible.


libgalois/include/galois/DynamicBitset.h
```diff
@@ -232,19 +230,14 @@ class DynamicBitSet {
     // if old_bit is 1, then atomically reset it
     while (((old_val & bit_offset) != 0) &&
            !bitvec[bit_index].compare_exchange_weak(
-               old_val, old_val & ~bit_offset, std::memory_order_relaxed))
+               old_val, old_val & ~bit_offset, std::memory_order_relaxed)) {
```
roshandathathri, line 233: Why not move move this and the above functions to the cpp files?

Just curious about the reason.

ddn0: I wasn't sure if some of these functions were intended to be inlined. There was a previous commit to move things from the cpp file to the headers, so I stuck with what is definitely performance safe: just moving the all the functions that used do_all/on_each.

@bozhiyou @l-hoang is it fine to move the rest of the functions to the cpp file?

l-hoang: I'm probably not one to speak on C++ issues w.r.t. performance, so anything I may have done in the past might have had no theoretical basis.

The reasons I keep implementations in headers are the following:
1) Templated functions can't be in .cpp anyways
2) Laziness

I've never thought about performance for better or worse other than how long things take to compile if you have too many things in a header.

In other words, I have no problem with it assuming there's no performance penalty (which I'm not sure would exist or not).

roshandathathri: @ddn0 Your logic is sound, as always. Inlineable or not is a good rubric.

I implemented those functions initially. I think it makes sense to continue keeping the simple test, set, reset functions in the header.

bozhiyou: I was the culprit of the [previous commit to move things from the cpp file to the headers](https://github.com/IntelligentSoftwareSystems/Galois/commit/ff8b72709801dc8a5d9de91340fa1ae99403ca0c#diff-45ff0e0acee99ca23a14c9809af5488b) but can't recall the reason as the commit comment doesn't explain either (what a bad practice I did!). One thing for sure is that this change was not due to performance issue - even till now I don't know how putting functions in headers or not affects being performance safe (and I'm curious to know the answer).

PR #295, opened 2020-07-30
Parameters useable query test, used in skip/limit

l-hoang:
Adds initial parameter support to the querying engine. Users pass in
parameters via the command line in the format \<paramname\>=\<paramvalue\>
via the -parameter command line option. They can choose to pass the
arguments in a comma separated fashion or individually via repeated
invocations of the command line argument.

Example:

-parameter=arg1=1 -parameter=arg2=2
-parameter=arg1=1,arg2=2.

The parameter map is available to the backend. Right now, they are only
used for skip/limit definitions. The idea is that eventually they will
be used during filtering when users define strings/literals.


lonestar/querying/cpu/querytest/querytest.cpp
```diff
@@ -47,6 +47,13 @@ static cll::opt<std::string>
                         "; takes precedence over query string"),
               cll::init(""));
 
+static cll::opt<std::string> param_file(
+    "paramFile",
+    cll::desc(
+        "File containing parameters to be used by queries at runtime; the file "
+        "contains <paramname>=<value> for each parameter on separate lines"),
+    cll::init(""));
```
roshandathathri, line 55: This is not a good way to take key,value pairs. 

A better way is to use cl:list to take multiple "key=value" strings from the command line. You can also take multiple "key,value" strings from the command line and let cl:CommaSeparated handle it for you.
https://llvm.org/docs/CommandLine.html
I have not tried either of those options, so don't know how they will work. Try it out.

libquery/src/AttributedGraph.cpp
```diff
@@ -458,6 +460,23 @@ void printArrayElement(const std::shared_ptr<arrow::DataType>& d_type,
     }
   }
 }
+
+std::optional<int> ReturnParameterAsInt(
+    const std::unordered_map<std::string, std::string>& parameter_map,
+    std::string parameter_name) {
+  // skip value is in a parameter
```
roshandathathri, line 467: How is this specific to skip?

libquery/src/AttributedGraph.cpp
```diff
@@ -458,6 +460,23 @@ void printArrayElement(const std::shared_ptr<arrow::DataType>& d_type,
...
+  auto param_pair = parameter_map.find(parameter_name);
+
+  if (param_pair != parameter_map.end()) {
+    return std::stoi(param_pair->second);
```
roshandathathri, line 471: So you're assuming the type is Int? This is fine for skip and limit but not sure whether similar functions can handle arbitrary types for attributes/properties.

lonestar/querying/cpu/querytest/querytest.cpp
l-hoang: will look into it

libquery/src/AttributedGraph.cpp
l-hoang: It isn't; leftover from some refactoring I did.

libquery/src/AttributedGraph.cpp
l-hoang: For the function, yes, since it's called Return As Int.

For skip/limit, for now, yes: technically it can be some arbitrary expression (e.g. $param + 3, or use some property from the graph), but the compiler is only supporting ints at the moment. I don't consider arbitrary expression support  a priority right now.

For arbitrary things, according to the spec a param can really only be a string or a literal (in other words, int or float); it shouldn't be too hard to deal with those I think.

roshandathathri: Sounds good.

l-hoang:
Thanks for pointing out cl::list @roshandathathri ; much cleaner now + more compatible with any Jupiter frontend I would think.

lonestar/querying/cpu/querytest/querytest.cpp
```diff
@@ -119,9 +126,28 @@ int main(int argc, char** argv) {
...
   galois::loadKatanaFormToCypherGraph(att_graph, filename);
   graphConstructTime.stop();
-
   att_graph.reportGraphStats();
 
+  // holds parameter values for cypher query passed in via command line
+  std::unordered_map<std::string, std::string> parameter_map;
+  // parameter parsing
+  for (unsigned i = 0; i < param_list.size(); i++) {
+    // TODO currently assumes input passed is perfect; need input checks
```
roshandathathri, line 135: What checks are you referring? Also "perfect" is vague. Do you mean "sanitized"?

l-hoang: Checks to make sure the format a user passes a parameter in is sane. Sanitized is a much better term; will change it.

Might as well add a check for an "equals" sign as well since that's really the only thing it expects; everything else I guess is fair game.

roshandathathri: Makes sense.

l-hoang: done, check the updated version

roshandathathri: Looks good

l-hoang:
going to assume Gurbinder will be fine with command line passing w.r.t. Jupytr; merging

PR #301, opened 2020-07-30
python: Add CI documentation build and artifact upload

arthurp:
This PR makes disables Python documentation build by default (so Sphinx isn't required by default) and setups up the CI to enable documentation build and upload an artifact with the built documentation.


python/docs/CMakeLists.txt
```diff
@@ -1,36 +1,38 @@
...
-        -q -b html
-        -c "${BINARY_BUILD_DIR}"
-        -d "${SPHINX_CACHE_DIR}"
-        "${CMAKE_CURRENT_SOURCE_DIR}"
-        "${SPHINX_HTML_DIR}"
-    COMMENT "Building HTML documentation with Sphinx")
-
-add_dependencies(python_doc shmem graphs timer atomic datastructures _loops property_graph)
\ No newline at end of file
+if(BUILD_DOCS)
```
ddn0, line 3: I think you can just write

```cmake
if(NOT BUILD_DOCS)
  return()
endif()

....
```

python/conda_recipe/build.sh
```diff
@@ -5,7 +5,20 @@ if ccache -V > /dev/null 2> /dev/null; then
     CMAKE_CCACHE_OPTION="-DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
     echo "Enabling CCache."
 fi
+
```
ddn0, line 8: Do we want a `set -e` or `set -eu -o pipefail`?

python/conda_recipe/build.sh
```diff
@@ -5,7 +5,20 @@ if ccache -V > /dev/null 2> /dev/null; then
...
 cd $SRC_DIR
 rm -rf _skbuild
-export GALOIS_CMAKE_ARGS="-DBUILD_LIBGALOIS=OFF -DCMAKE_BUILD_TYPE=Release -DUSE_ARCH=none $CMAKE_CCACHE_OPTION"
+
+export GALOIS_CMAKE_ARGS="-DBUILD_LIBGALOIS=OFF -DCMAKE_BUILD_TYPE=Release -DUSE_ARCH=none $CMAKE_CCACHE_OPTION $CMAKE_DOCS_OPTION"
 $PYTHON setup.py --verbose build_ext --verbose install -- -- -j2
+
+if [ -n "$GALOIS_DOCS_OUTPUT" ]; then
+    echo "Exporting documentation to $GALOIS_DOCS_OUTPUT"
+    mkdir -p "$GALOIS_DOCS_OUTPUT"
```
ddn0, line 22: I don't know if this really matters but `$GALOIS_DOCS_OUTPUT` needs to be an absolute path because of `cd $SRC_DIR`. Probably doesn't matter to fix this though. I just thought I'd point it out.

PR #317, opened 2020-08-04
Class for holding bit to label mappings

l-hoang:
BitLabelSchema is a class that holds a mapping from a bit to its
corresponding node/edge label. This information used to be directly
contained in AttributedGraph, but it has been put into its own class.
This removes a circular dependency between QueryIR and AttributedGraph
as QueryIR needs this information to construct the query graph, and
before this change, AttributedGraph had to be included as a header to
get this information.

In addition, this is a step towards decoupling the information that
differentiates AttributedGraph from a regular PropertyFileGraph; this
class can be used in the future with PropertyGraph when the engine
eventually moves to start using that as its main graph representation
instead of LC_CSR.

At this moment in time as a matter of convenience AttributedGraph still
contains a pointer to the BitLabelSchema so I don't have to pass it to
every function that uses it, but in the future, the idea is to pass this
object only when its required rather than having it remain as a pointer
in the graph class.

This commit also includes a few other things like using pointers for
modifiable arguments per Katana convention and using std::optional
instead of a std::pair with a boolean flag.


libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -43,24 +38,23 @@ class AttributedGraph {
...
-  std::unordered_map<std::string, uint32_t> node_label_ids;
-  //! Given an edge label i, get its name
-  std::vector<std::string> edge_label_names;
-  //! Given an edge name, get its numerical index i
-  std::unordered_map<std::string, uint32_t> edge_label_ids;
+  // TODO in the future, this schema class may not necessarily be tied to the
+  // PropertyGraph class; this is just convenience for now so I don't have to
+  // pass a reference to the bit label schema to every function that uses it
+  //! Bit label schema containing mappings from bits to labels for nodes/edges
+  const BitLabelSchema* bit_label_schema{nullptr};
```
roshandathathri, line 45: I would prefer if this is a const reference.

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -43,24 +38,23 @@ class AttributedGraph {
...
-  std::vector<std::string> node_label_names;
-  //! Given a node name, get its numerical index i
-  std::unordered_map<std::string, uint32_t> node_label_ids;
-  //! Given an edge label i, get its name
-  std::vector<std::string> edge_label_names;
-  //! Given an edge name, get its numerical index i
-  std::unordered_map<std::string, uint32_t> edge_label_ids;
+  // TODO in the future, this schema class may not necessarily be tied to the
+  // PropertyGraph class; this is just convenience for now so I don't have to
+  // pass a reference to the bit label schema to every function that uses it
```
roshandathathri, line 43: I don't agree with "in the future". The schema is fixed/maintained by the graph, so it doesn't make sense to pass it to the graph's member functions (other than the constructor).

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -74,15 +68,14 @@ class AttributedGraph {
...
-  //! @param numNodeLabels Number of node labels to allocate memory for
-  //! @param numEdgeLabels Number of edge labels to allocate memory for
-  void AllocateGraph(size_t numNodes, size_t numEdges, size_t numNodeLabels,
-                     size_t numEdgeLabels);
+  //! @param num_nodes Number of nodes to allocate memory for
+  //! @param num_edges Number of edges to allocate memory for
+  //! @param _bit_label_schema Pointer to the bit label schema mapping bits to
+  //! labels
+  void InitializeGraph(size_t num_nodes, size_t num_edges,
+                       BitLabelSchema* _bit_label_schema);
```
roshandathathri, line 78: Why is this not a reference?

libquery/include/querying/BitLabelSchema.h
```diff
@@ -0,0 +1,86 @@
...
+  std::unordered_map<std::string, uint32_t> node_label_ids;
+  //! Given an edge label i, get its name
+  std::vector<std::string> edge_label_names;
+  //! Given an edge name, get its numerical index i
+  std::unordered_map<std::string, uint32_t> edge_label_ids;
+
+public:
+  //! Allocates space for the data structures of this class. This must be
+  //! called before setting any new node metadata.
+  void AllocateMemory(size_t num_node_labels, size_t num_edge_labels);
```
roshandathathri, line 26: Why not make this a constructor? If there is a need to change number of node/edge labels dynamically, then have separate functions to increase number of node/edge labels.

libquery/include/querying/BitLabelSchema.h
```diff
@@ -0,0 +1,86 @@
...
+  std::optional<std::pair<uint32_t, uint32_t>>
+  GetEdgeLabelMask(const std::string edge_label) const;
+
+  //! Return the node labels in a string
+  std::string AllNodeLabelsAsString() const;
+
+  //! Return the edge labels in a string
+  std::string AllEdgeLabelsAsString() const;
+
+  //! Given a bitmap, return a string containing the node labels represented by
```
roshandathathri, line 70: Semi-colon separated string? Or just a concatenated string?

libquery/include/querying/BitLabelSchema.h
```diff
@@ -0,0 +1,86 @@
...
+  //! the bitmap
+  std::string GetNodeLabelStringFromBitmap(uint32_t bitmap) const;
+
+  //! Given a bitmap, return a string containing the edge labels represented by
```
roshandathathri, line 74: Ditto

roshandathathri:
> using pointers for
> modifiable arguments per Katana convention

First, where is this convention noted?
Second, I dislike this convention. Modifiable and read-only reference arguments can be distinguished by using non-const and const reference arguments. Why is there a need to use a pointer?

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: Same, but that would require initialization of the reference at runtime (so constructor would need to take the reference). If we can guarantee that the schema object is created before the graph object then that design is fine.

libquery/include/querying/BitLabelSchema.h
l-hoang: I do not have this information available at the point this object is declared. One option would be to return the object from the LoadKatanaGraph function so that I only create it in that function instead of passing a pointer to it.

libquery/include/querying/BitLabelSchema.h
l-hoang: Concatenated string where labels are separated by colons (like Cypher syntax). I'll add that to the comment.

libquery/include/querying/BitLabelSchema.h
l-hoang: ditto reply

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: My thought process for this was that I didn't want to have to require its constructor to have the Schema object, but you are right in that it probably makes the most sense. Will do a bit of redesign.

l-hoang:
> First, where is this convention noted?

https://github.com/KatanaGraph/katana/blob/master/CONTRIBUTING.md

> One exception (so far) is that when passing arguments to functions, prefer pointers to non-const references (which used to be a part of that guide).

This implies that non-modifiable arguments are const references while modifiable arguments are pointers according to my understanding of this.
 
> Second, I dislike this convention. 

Ditto.

> Modifiable and read-only reference arguments can be distinguished by using non-const and const reference arguments. Why is there a need to use a pointer?

My guess is the syntax makes it obvious that we're making modifications.

If you use a pointer, function calls/field accesses use ->.
With a reference, it's ".".

Seeing this in code makes it easier to distinguish between use cases.

libquery/include/galois/graphs/AttributedGraph.h
l-hoang: See comment about initialization in the AllocateMemory comment. Also due to Katana preference.

libquery/include/querying/BitLabelSchema.h
l-hoang: Looked into this option; complicates matters because AttributedGraph needs to be initialized before LoadKatana is called, and the new design requires that AttributedGraph have a bit label schema object at init time to initialize the const reference (rather than the pointer that was originally there).

roshandathathri: Thanks for checking.

PR #322, opened 2020-08-05
python: Use numba jitclasses to store closures. Fix memory leak.

arthurp:
This still requires hacks and probably depends on internals of numba, but it doesn't leak memory and it is less of a hack.


arthurp:
> Is there any performance overhead expected?

There may be yes. In fact, I just realized it might be significant. There is an additional atomic instruction on the operator call path due to how memory management is handled in Numba. I'll think about this. It should be possible to move the atomics out of the hot/parallel path. Or at least I hope it is.

My opinion is that we should merge this ASAP anyway and deal with the performance along with everything else, since this fixes a memory leak which is arguably a correctness problem.

gurbinder533:
> My opinion is that we should merge this ASAP anyway and deal with the performance along with everything else, since this fixes a memory leak which is arguably a correctness problem.

I agree. Let's merge this then and re-evaluate before we start patching up performance.

arthurp:
BTW, I had a realization. I think the atomic instruction I'm worried about gets optimized out. I'll verify that when I work on performance.

PR #323, opened 2020-08-05
CI: Add python format checking

arthurp:


arthurp:
@ddn0, please check the CI calls and stuff.

@gurbinder533, please make sure you are OK with the python formatting style this enforces. This is up for discussion.

.github/workflows/conda.yaml
```diff
@@ -1,4 +1,4 @@
-name: Conda Package CI
+name: Conda Package CI and Python Formatting
```
ddn0, line 1: maybe it is time to rename this file `python.yaml`?

.github/workflows/conda.yaml
```diff
@@ -135,3 +135,71 @@ jobs:
...
+        bash "$HOME/.cache/miniconda/miniconda.sh" -f -b -p $HOME/miniconda
+        export PATH=$HOME/miniconda/bin:$PATH
+        source activate
+        conda install -y -q -c conda-forge black pylint
+    - name: Check formating
+      run: |
+        export PATH=$HOME/miniconda/bin:$PATH
+        source activate
+        ./scripts/py_fmt.sh python tests
+#    - name: Check for lint
```
ddn0, line 201: Questions:
- The linting is disabled because of python extension modules?
- We don't have enough files to justify running the linter on just vanilla python files?

arthurp: The issue is that running the linter on the python files doesn't work well because:

* Many modules import python extension modules and pylint complains about missing imported values.
* There is a decorator which changes the number of arguments a function expects and pylint complains about each call to a function using that decorator.

I'll investigate disabling argument checking on any function with that decorator, and I can just disable import checking entirely (I don't see any other option). I guess there would still be lint useful lint checks left even with those disabled. I'll give it a try.

scripts/py_lint.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
ddn0, line 1: some preference for these new scripts to follow the naming the existing CI scripts, `check_....`.

.github/workflows/conda.yaml
arthurp: There is a really large number of lint warnings, but I did go through and clean up some of them and I did figure out how to prevent the most obviously incorrect warnings. You can look at the output to see how many errors there are right now. I'll come back to this as I have time.

ddn0: FWIW, we can also merge this PR and keep the pylint check not required to merge and address the rest of the pylint errors in a separate commit.

arthurp: True. But my concern is that it will be really irritating that every PR and every commit will appear with a red X since even if the result isn't required it is reported as an error.

arthurp:
@ddn0 I finally went through and eliminated all the pylint errors on my local machine but when I uploaded the CI detected more. Have you ever encountered this issue? I have no idea why it's happening, the versions of pylint are the same in CI and on my local machine and as far as I can tell the whole configuration is basically the same.

insertinterestingnamehere:
My delayed two cents here, IIRC, pylint tends to have a lot of false positives and unhelpful warnings. SciPy has a CI build similar to what you're trying to set up here, but they use pyflakes and pycodestyle instead. See https://github.com/scipy/scipy/blob/master/.travis.yml#L28-L42 for a working example.

arthurp:
Given the problems with pylint and @insertinterestingnamehere's valid point about false positives in pylint and other potentially better tools, I'm going to shelve the pylint stuff. I will create a new PR other people can start from it they want. I will not have time I should use on this given how soon my internship is ending.

I'm going to merge this PR with just the black-based formatting/indentation.

PR #332, opened 2020-08-07
A bunch of header files to guide the development of the execution plan

bowu:
The design document is at: https://docs.google.com/document/d/106mNJR-DG17P9RMAf3DLoIFu67MoBmNlN-Oh_NPq_5w/edit?usp=sharing


libquery/include/querying/ExpressionTree.h
```diff
@@ -0,0 +1,122 @@
+enum BinaryTreeOperator {
```
l-hoang, line 1: I would remove this file; it will be added in the PR I'm currently working on.

bowu:
> 2 requests, then some comments about overall design.
> 
> 1. Remove the external/cub changes that seem to have gotten picked up with this PR.
> 2. Remove libquery/include/querying/ExpressionTree.h; I'll PR that file once it's done (currently working on it).

Will follow.

> Now for the comments:
> 
> Design comments:
> 
> * Right now the operators are named after Cypher keywords mostly. For generality's sake, it probably makes more sense to avoid tying the functionality to Cypher. e.g., skip, limit, and order by can probably be merged into some kind of "table manipulation" op, esp. since I don't think having skip/limit nodes by themselves really makes sense since they're relatively minor operations.

We will have optimizers to optimize out some operators, but I think having those operators makes at least debugging easier.

> * Similar to the above point, I noticed there's no RETURN op, but then I realized that return/with should probably be a single "select/table creation" op.
> * I think aggregation being it's own operator is fine, but what about cases where we can do aggregation as an optimization in the match? This may be a case of premature optimization though.

Again, some optimizer will determine whether/how we should transform the OpGraph to perform this optimization.

> * In general, I feel like the MATCH op can do a _lot_ of things that the other ops are defined for. I'm not sure how to think about the MATCH op individually given that that's the case or how it would theoretically compose with the other ops.

The MATCH op will indeed do a lot of things, but semantically speaking it should just find all instances and populate a table based on them. Whether it should perform tasks of other ops depends on our optimizers.

> Code comments:
> 
> * We're using PascalCase for function names and snake_case for variable names in new Katana code.

Will follow.

> * I'm not too sure about the current directory structure, though for now it's fine. I feel like the querying subdirectory might end up with too many things it in.

Good point. We can revisit this later.

roshandathathri:
@bowu There are still some unrelated submodule changes in the PR (submodules cub and moderngpu are deleted; external bliss files are deleted). Please undo those changes and push only the files that you changed.

libquery/include/querying/ops/Operator.h
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  size_t num_ready_inputs_{0};
+
+  uint32_t id_;
+
+  OpType type_;
+
+  std::chrono::duration<float> duration_;
```
l-hoang, line 99: CI complains about this field not being used

lonestar/querying/cpu/CMakeLists.txt
```diff
@@ -1 +1,2 @@
 add_subdirectory(querytest)
+add_subdirectory(ExecutionPlanTest)
```
l-hoang, line 2: Not sure what our convention for directory names/tests is, but I don't think it's Pascal case for tests.

lonestar/querying/cpu/ExecutionPlanTest/ExecutionPlanTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+static cll::opt<std::string> graphfile("graphfile", cll::desc("path to Katana Form metafile"),
+                                   cll::init(""),
+                                   cll::Required);
+
+static cll::opt<std::string> query("query", cll::desc("Cypher query"),
+                                   cll::init(""),
+                                   cll::Required);
+
+static cll::opt<uint32_t> num_pages("numPages",
```
l-hoang, line 30: Unused argument

PR #347, opened 2020-08-11
Converter: Refactor arrow and topology building code

patrickkenney9801:
Refactors arrow and topology code into a PropertyGraphBuilder class so that converters for different sources do not need to handles arrow/topology logic themselves


tools/graph-convert/tests/graph-properties-convert-test.cpp
```diff
@@ -1100,49 +1099,49 @@ galois::GraphComponents GenerateAndConvertBson(size_t chunk_size) {
...
 void VerifyMongodbSet(const galois::GraphComponents& graph) {
 #ifndef GALOIS_MONGOC_FOUND
   (void)graph;
   return;
-#endif
-  GALOIS_ASSERT(graph.node_properties->num_columns() == 2);
-  GALOIS_ASSERT(graph.node_labels->num_columns() == 1);
-  GALOIS_ASSERT(graph.edge_properties->num_columns() == 1);
-  GALOIS_ASSERT(graph.edge_types->num_columns() == 1);
+#else
```
ddn0, line 1120: some preference for simplifying the conditional preprocessing by pulling out mongodb stuff to a separate file. In any event, adding something like
```cpp
int main(...) {
...

switch (file_type) {
...
#ifdef GALOIS_MONGOC_FOUND
  case galois::SourceDatabase::kMongodb:
    ...
#endif
  default:
    GALOIS_LOG_FATAL(...);
}
...

}
```

will ensure that we fail the test if we try to run a mongodb test when we are built without mongodb rather than silently succeeding.

ddn0:
Much cleaner than before. Thanks!

tools/graph-convert/tests/graph-properties-convert-test.cpp
patrickkenney9801: the point of adding it so that mongodb did not build was so that we could pass python tests...

tools/graph-convert/graph-properties-convert.cpp
```diff
@@ -284,6 +289,176 @@ void WriteFalseStats(const std::vector<ArrowArrays>& table,
...
+  }
+  case arrow::Type::TIMESTAMP: {
+    auto tb  = std::static_pointer_cast<arrow::TimestampBuilder>(array);
+    auto res = ResolveValue(ImportDataType::kTimestampMilli, is_list);
+    if (res.type != ImportDataType::kUnsupported) {
+      st = tb->Append(std::get<int64_t>(res.value));
+    }
+    break;
+  }
+  // for now uint8_t is an alias for a struct
```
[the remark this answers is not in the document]
patrickkenney9801: this should be changed I agree, for now it is really a boolean bitmap for whether an index in a struct array is null or not, I just did not use boolean since we were moving away from that
right now I do not think we have very strong type support on the querying side anyways and I am not sure how they would want to deal with structs when they are ready
if we disallow users from getting uint8_t or int8_t values then we could continue using them for booleans as we already are, of course this stuff can always be changed

tools/graph-convert/tests/graph-properties-convert-test.cpp
ddn0: You could just add something to tools/.../tests/CMakeLists.txt like:
```cmake
if(mongoc-1.0_FOUND)
  add_test(NAME convert-properties-mongodb ...)
else()
  message(STATUS "Skipping mongodb tests")
endif()
```

More work, I know, but in the off chance someone has a `graph-properties-convert-test` program and runs it with `--mongodb` manually without mongodb support, I think would be better to report an error than to silently return success.

patrickkenney9801: so I just add that message? that would not be much work?

PR #350, opened 2020-08-11
tsuba: don't check the ec2 metadata service

### LINKED ISSUE #349: Revisit S3 credential provider chain after aws-cpp-sdk is updated in conan.
We should be able to go back to using the default provider chain once the version provided is updated.

tylershunt:
Commit of shame.

Found the real culprit an its an http request if credential selection falls through to the EC2 MetaData service.

Closes #349


ddn0:
Commit of learning!

Since our standard deployment platform is AWS and people naturally might want to use the metadata service in that case and we are introducing a non-standard environment variable (`AWS_EC2_METADATA_ENABLED`), would it make sense to just recommend to people not running in AWS to set some dummy credentials or something?

I suppose if you wanted to be really nice you could wrap the metadata provider with a warning message if it doesn't return successfully to suggest to the user to set some credentials.

tylershunt:
> Since our standard deployment platform is AWS and people naturally might want to use the metadata service in that case and we are introducing a non-standard environment variable (`AWS_EC2_METADATA_ENABLED`), would it make sense to just recommend to people not running in AWS to set some dummy credentials or something?

That's a good point. Yeah leaving the default and printing a warning instead makes more sense.

tylershunt:
Updated to preserve the default chain with a warning to suggest options to improve performance when no credentials are found.

ddn0:
Updated the PR to close #349.

tylershunt:
Added @bowu and @l-hoang as reviewers mainly just because they seem the most interested in this issue.

PR #351, opened 2020-08-11
Remove global load for external function calls

arthurp:
Initial unscientific measurements shows this to produce ~7% performance improvement on pagerank.

The downside is that it further couples us to the internals of Numba, so it will make numba upgrades potentially cause problems. Unit testing will hopefully catch any problems an upgrade would produce.


python/galois/numba/galois.py
```diff
@@ -22,10 +22,11 @@ def impl(self):
...
 @overload_method(galois.property_graph.PropertyGraph_numba_wrapper.Type, "edges")
 def overload_edges(self, n):
-    def impl(self, n):
-        if n == 0:
-            prev = 0
-        else:
-            prev = self.edge_index(n-1)
-        return range(prev, self.edge_index(n))
-    return impl
+    if isinstance(n, types.Integer) and not n.signed:
```
gurbinder533, line 25: What happens or what does it return in case "if" fails?

python/galois/numba/utils.py
```diff
@@ -1,14 +1,12 @@
 from llvmlite import ir
 
 
-def call_raw_function_pointer(func_ptr, function_type, args, c):
-    function_type = ir.FunctionType(ir.PointerType(ir.IntType(8)),
-                                    (ir.PointerType(ir.IntType(8)),))
+def call_raw_function_pointer(func_ptr, function_type, args, builder: ir.IRBuilder):
     ptr = ir.Constant(ir.IntType(64), func_ptr).inttoptr(ir.PointerType(function_type))
     # HACK: Add a field to ptr which is expected by builder.call based on the
     #  assumption that the function is a normal Function.
```
gurbinder533, line 7: Are there any side-effects of this hack? What does "normal" function mean here?

arthurp: I think there is a way to avoid this hack entirely. Give me an hour. I'll push an update.

python/galois/numba/galois.py
arthurp: The function returns None (python semantics mean any path without a return implicitly returns None) which is interpreted as the function call not type checking, so numba will try other implementations or raise a type error.

python/galois/numba/utils.py
arthurp: This should be fixed now.

PR #354, opened 2020-08-12
MatchOp construction for Query Operation Graph

l-hoang:
MatchOp construction for Query Operation Graph

This commit adds support for the construction of a Match node in the 
operation graph. It adds the hooks for generating the rest of the 
operation graph nodes as well (abstract WITH and abstract RETURN); the 
actual construction of these operations will be handled at a later date.

The MatchOp is constructed by constructing the query graph from a set of
parsed MATCH clauses from a Cypher AST. One query graph is constructed 
for every scope (scopes are delimited by WITH clauses in the AST). 
FilterOps should be constructed from the set of MATCH clauses as well, 
but that will be handled once expression tree support for the AST parser
is in.

QueryIr.h/cpp are now deprecated as most of the functionality will be 
moved to the OpGraph. They have been marked as such in the files. They 
will be removed once end-to-end execution using an OpGraph works as they
are still needed to run the querytest application.


libquery/src/Operator.cpp
```diff
@@ -0,0 +1,8 @@
+#include "querying/ops/Operator.h"
+
+namespace galois {
+//! Enum to name map.  Must make sure the order in opNames matches that in OpType
+const std::string Operator::op_names[] = {"MatchOp",   "FilterOp",  "WithOp",
+                                          "GroupByOp", "OrderByOp", "LimitOp",
+                                          "SkipOp",    "ReturnOp"};
+}
```
bowu, line 8: Put this in the header file to reduce the risk of messing up the order. Do this:
static const constexpr char* const op_names[] = {"MatchOp", .. }

libquery/src/OpGraphBuilder.cpp
```diff
@@ -0,0 +1,167 @@
...
+      segments.emplace_back();
+      // save data
+      segments.back().segment_type = kASTMatch;
+      segments.back().child_start = match_start;
+      segments.back().child_end = i;
+      // reset
+      match_start = static_cast<size_t>(-1);
+    }
+
+    CypherASTSegmentType segment_type;
```
bowu, line 83: We have a different understanding of what a segment is. The boundary of a segment is determined by one (for the first and last segments) or two WITH statements (a segment is the AST in between). The benefit of having the segment concept is to more easily handle WITH. Therefore, you may change the workflow as follows. First, divide the query into segments demarcated by WITH statements. Second, build a scoping operator at the beginning of each segment to properly scope. Third, build an sub-OpGraph for each segment, which is what you are trying to do in this PR. Fourth, glue the sub-OpGraphs to form the final OpGraph.

libquery/src/Operator.cpp
l-hoang: It was originally in the header file, but there were issues with multiple definitions since the operator header appears in many places. (the header guard does not fix this according to some stack overflow post I read)

bowu: Use the statement in my comment. It works. I neglected constexpr previously.

libquery/src/OpGraphBuilder.cpp
roshandathathri: Agree with Bo. 

SegmentType is a misnomer here. Maybe you mean OpType?

l-hoang: Yes, I think "segment" is causing some confusion. I'll think of a better name for it.

bowu: @roshandathathri Loc and I believe that supporting WITH shouldn't be difficult in the current design if we embed the scoping information in the Table class.

libquery/include/querying/QueryServer.h
```diff
@@ -20,9 +20,12 @@ class QueryServer {
 
   Table RunCypherQuery(std::string graph_file_name, std::string query_str) {
     std::shared_ptr<AttributedGraph> graph = LoadPropertyGraph(graph_file_name);
-    std::shared_ptr<ExecutionPlan> plan    = cache_.GetPlan(query_str);
+    // TODO(bowu) plan/query graph is tied to some schema; query_str is not
+    // enough, need graph schema to create it; tie graph schema to cache as well
```
roshandathathri, line 24: "schema" is overloaded. I think you mean bit_label_schema here. We are not planning to use any other graph schema until much later.

libquery/include/querying/OpGraphBuilder.h
```diff
@@ -1,30 +1,63 @@
...
+  //! Adds a new operation to an op graph, adds the new operation as a child to
+  //! the last operation, and updates the pointer to the last operation that
+  //! this builder built (last_op_)
+  void AddOpToOpGraph(OpGraph* built_op_graph,
+                      std::shared_ptr<Operator> new_op);
+
+  //! Constructs and adds an abstract match operator to a given operation graph.
+  //! This will involve at least a MATCH node and possibly a FILTER node
+  //! to operation graph.
+  void AddMatchOp(OpGraph* built_op_graph,
```
roshandathathri, line 58: It seems a bit weird to pass around op_graph in these two functions. I think op_graph is expected to be returned and used even after Builder object is destroyed. If so, the member variable can be a unique_ptr. Unless there are multiple op_graph values that can be passed to these functions, it's better to use a member variable.

l-hoang: > Regarding the design, my main comment is that either OpGraphBuilder and CypherCompiler should be merged or they should have a more well-defined interface between them. I think the latter is a better choice because CypherCompiler can be replaced with say GremlinCompiler to support other languages.

Oh, I see what you mean now. MatchOp shouldn't have to know how CypherCompiler works to create the MatchOp, so there's still some entanglement there.

I'll think about how to fix that. The way it's being done now maximizes the possible code reuse since we're on a tight self-imposed deadline.

l-hoang:
> Regarding the design, my main comment is that either OpGraphBuilder and CypherCompiler should be merged or they should have a more well-defined interface between them. I think the latter is a better choice because CypherCompiler can be replaced with say GremlinCompiler to support other languages.

Oh, I see what you mean now. MatchOp shouldn't have to know how CypherCompiler works to create the MatchOp, so there's still some entanglement there.

I'll think about how to fix that. The way it's being done now maximizes the possible code reuse since we're on a tight self-imposed deadline.

libquery/include/querying/OpGraphBuilder.h
l-hoang: I'm not exactly sure how the unique_ptr semantics work if I return a member variable as the return value for the Build function which is why it's currently a local variable in the Build function. This is something I can look into when I add Return op support which should be coming up soon.

roshandathathri: It should work but you can try it out later.

l-hoang:
I'll resolve Roshan's 2 comments in the next PR for ReturnOp construction.

PR #359, opened 2020-08-14
Leave a trail of previous RDG meta files

witchel:
Leave our metadata trail of breadcrumbs.  Make RDGMeta a public struct.  Set the stage for garbage collection, but I'm committing this part first.  Also, send file and line information when crash testing.


libsupport/src/FileSystem.cpp
```diff
@@ -74,12 +74,22 @@ galois::Result<std::string> galois::NewPath(const std::string& dir,
   return p.append(1, kSepChar).append(name);
 }
 
+// TODO: These functions do not properly deal with non-canonical path
+//  names, e.g., multiple '/' in a row
```
tylershunt, line 78: Should probably be `TODO(witchel)` or maybe `TODO(scober)`

libsupport/src/FileSystem.cpp
```diff
@@ -74,12 +74,22 @@ galois::Result<std::string> galois::NewPath(const std::string& dir,
...
   return path.substr(last_slash + 1);
 }
+
+galois::Result<std::string> galois::ExtractDirName(const std::string& path) {
+  size_t last_slash = path.find_last_of(kSepChar, std::string::npos);
+  if (last_slash == std::string::npos) {
+    return ErrorCode::InvalidArgument;
+  }
+  return path.substr(0, last_slash);
+}
```
tylershunt, line 95: The behavior here is to return `some/long/path` for an input like `/some/long/path/` that end in  `kSepChar`? I only ask because `dirname` would return `some/long` in that case.

libtsuba/include/tsuba/FaultTest.h
```diff
@@ -23,7 +23,8 @@ GALOIS_EXPORT void FaultTestInit(FaultMode mode         = FaultMode::None,
 GALOIS_EXPORT void FaultTestReport();
 
 // PullThePlug (virtually) Compile this out if NDEBUG?
-void PtP(FaultSensitivity sensitivity = FaultSensitivity::Normal);
+void PtP(const char* file, int line,
+         FaultSensitivity sensitivity = FaultSensitivity::Normal);
```
tylershunt, line 27: These are getting verbose; we might think about adding a macro like:
`#DEFINE TSUBA_PTP(...) ::tsuba::internal::PtP(__FILE__, __LINE__, ##__VA_ARGS__)`

witchel: I agree it is getting ugly.  For some reason, I thought we had a "ban" on cpp, but I'm happy to turn this into a macro.

libsupport/src/FileSystem.cpp
witchel: Sure, I can do this.  I'll point out that ExtractFileName would return the empty string if you send it  `/some/long/path/`

tylershunt: No real preference just wondering about the semantics. `basename /some/long/path/` returns `path` so maybe we should change there too?

libtsuba/include/tsuba/FaultTest.h
tylershunt: Yeah I think we want to avoid cpp macros in general, but for things involving  `__FILE__` and `__LINE__` I don't know of anything better.

libsupport/src/FileSystem.cpp
witchel: This code is ugly.  Maybe there is a better way?

PR #363, opened 2020-08-14
Add mysqlclient dependency and mysql converter to katana form

patrickkenney9801:
This allows schema mapping support in the same manner as mongodb but safer due to the better type safety of mysql


tools/graph-convert/graph-properties-convert-graphml.cpp
```diff
@@ -169,7 +161,10 @@ std::optional<std::vector<T>> ParseNumberList(std::string raw_list) {
   boost::split(elems, raw_list, boost::is_any_of(","));
 
   for (std::string s : elems) {
-    list.emplace_back(boost::lexical_cast<T>(s));
+    try {
+      list.emplace_back(boost::lexical_cast<T>(s));
+    } catch (const boost::bad_lexical_cast&) {
```
ddn0, line 166: print a warning once?

tools/graph-convert/graph-properties-convert-mysql.cpp
```diff
@@ -0,0 +1,884 @@
...
+  while ((row = mysql_fetch_row(tables.res))) {
+    auto lengths = mysql_fetch_lengths(tables.res);
+    for (size_t i = 0; i < num_fields; i++) {
+      table_names.emplace_back(std::string(row[i], lengths[i]));
+    }
+  }
+  return table_names;
+}
+
+/*
```
ddn0, line 228: ???

patrickkenney9801: this function works, I changed my implementation so it was no longer necessary but thought it might be useful with user input later, just forgot to delete it.

PR #371, opened 2020-08-17
libquery refactoring: .cpp, docs, constructors

l-hoang:
Moves the implementation of functions to a cpp file, adds more
documentation, adds copy/move constructor declarations, and removes the
declaration of constructors from subclasses of Operator.


libquery/include/querying/ops/Operator.h
```diff
@@ -45,45 +46,40 @@ class Operator : public std::enable_shared_from_this<Operator> {
...
   virtual bool SanityCheck() = 0;
 
-  void AddParent(Operator* parent) {
-    parents_.emplace_back(parent);
-    parent->AddChild(this);
-    inputs_.emplace_back(nullptr);
-  }
+  //! Adds this node as a child to the passed in parent and add the parent
+  //! relation to this node as well.
+  void AddParent(Operator* parent);
```
roshandathathri, line 51: It's better to have such small functions in the header because it will allow them to be inlined.

libquery/include/querying/ops/Operator.h
```diff
@@ -45,45 +46,40 @@ class Operator : public std::enable_shared_from_this<Operator> {
...
-                       inputs_.size());
-    }
-    inputs_[num_ready_inputs_++] = d;
-  }
+  inline std::string GetOpTypeStr() const { return op_names[type_]; }
 
-  //! Whether the operator is ready to compute
-  bool IsReady() { return num_ready_inputs_ == inputs_.size(); }
+  //! Accept a new input that is ready for processing into this operator
+  void TakeInput(std::shared_ptr<InterData> d);
```
roshandathathri, line 59: Ditto

libquery/include/querying/ops/Operator.h
```diff
@@ -45,45 +46,40 @@ class Operator : public std::enable_shared_from_this<Operator> {
...
-    std::fill(inputs_.begin(), inputs_.end(), nullptr);
-    output_ = nullptr;
-  }
+  //! Return whether or not the operator is ready to compute; it is ready
+  //! when all its inputs have been accepted
+  inline bool IsReady() const { return num_ready_inputs_ == inputs_.size(); }
 
-  void SetDataGraph(graphs::AttributedGraph* dataGraph) {
+  //! Clear inputs and ouputs for the next round of computation
+  void Clear();
```
roshandathathri, line 66: Ditto

libquery/include/querying/ops/Operator.h
l-hoang: My criteria for moving something to the .cpp was if it was more than 1-2 lines or used if conditions; the idea there is that any declaration in the header file should not be more than one line if possible (so it's easier for someone new to code to see all functions at a glance since it's just line scanning at that point).

l-hoang: Looks like this is a good way to do it as well;
https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

Continuing through the code I realized that there are a lot of implementations in headers, so going to hold off on merging this for now and will do it as one big merge of refactoring.

roshandathathri: Makes sense. That's useful. Thanks for sharing.

This only applies if the definition is compiled into a static library (which is the case here). If it is a shared/dynamic library, a function definition not in the header cannot be inlined.

libquery/include/querying/ops/Operator.h
```diff
@@ -45,45 +46,40 @@ class Operator : public std::enable_shared_from_this<Operator> {
...
-    parent->AddChild(this);
-    inputs_.emplace_back(nullptr);
-  }
+  //! Adds this node as a child to the passed in parent and add the parent
+  //! relation to this node as well.
+  void AddParent(Operator* parent);
 
   //! Get operator type
-  OpType GetOpType() { return type_; }
+  inline OpType GetOpType() const { return type_; }
```
ddn0, line 54: `inline` is superfluous when a member function is defined inside the class definition: https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

libquery/include/querying/ops/Operator.h
```diff
@@ -6,13 +6,14 @@
 #include <chrono>
 
 #include "querying/InterData.h"
-#include "galois/Logging.h"
 #include "galois/graphs/AttributedGraph.h"
 
 namespace galois {
 
 class Operator : public std::enable_shared_from_this<Operator> {
```
ddn0, line 13: is `std::enable_shared_from_this` still needed?

l-hoang: @bowu

libquery/include/querying/ops/Operator.h
ddn0: In general, unless you need it for linking reasons, like declaring a free function in a header file, it is best to avoid using `inline` explicitly. Either the method definition is short, in which case you'd put it in the class definition anyway, or you should just let the compiler figure this out.

l-hoang: So even for the short functions in the cpp file I should avoid using inline? This says something else.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inline

ddn0: Yes. Unless you actually have observed that it is a performance issue (i.e., time critical), I'd avoid using inline.

l-hoang: Cool, thanks for the pointers!

libquery/include/querying/ops/Operator.h
bowu: No, it's not. @l-hoang Please push this commit. I've modified Operator.h and will PR soon.

l-hoang: you can go ahead and PR yours; this will take a while (going to be editing more than just this) and I can resolve conflicts as necessary

l-hoang:
Question @ddn0 

The operator class disables its move/copy constructors, so implicitly to my understanding its subclasses also don't have move/copy constructors. Is it better to redeclare the deleted move/copy constructors in the subclasses?

libquery/include/querying/QueryServer.h
```diff
@@ -12,51 +12,26 @@ namespace galois {
...
-    galois::LoadKatanaFormToCypherGraph(&(*new_graph), filename);
-    graphs::AttributedGraph* result = new_graph.get();
-    auto t = std::make_pair(filename, std::move(new_graph));
-    graphs_.emplace_back(std::move(t));
-    return result;
-  }
+  //! Loads a particular property graph and saves it to a set of loaded graphs
+  graphs::AttributedGraph* LoadPropertyGraph(std::string filename);
 
+  // TODO is this better as an unordered_map?
```
roshandathathri, line 34: For programmability, it's better to use an unordered_map. The number of elements in it is expected to be small, so performance isn't a concern.

l-hoang: Changed it into an unordered map while I was fixing the shared build.

libquery/include/querying/ops/WithReturnBaseOp.h
```diff
@@ -9,12 +9,7 @@ namespace galois {
 //! properties that need to be returned in a table
 class WithReturnBaseOp : public Operator {
 public:
-  // default constructor doesn't exist, needs return properties
-  WithReturnBaseOp()  = delete;
   ~WithReturnBaseOp() = default;
```
ddn0, line 12: Do we even need `~T() = default`?

l-hoang: I think the compiler complains about not having it IIRC since the destructor in the Operator superclass is virtual.

One option is to remove the virtual destructor completely and declare it as a default; I think that should work.

l-hoang: Note for those reading later: the answer is that you don't need ~T() even if the superclass's destructor is virtual = default.

PR #374, opened 2020-08-18
Operator removal

bowu:
This commit supports removing ops from OpGraph, which is needed for rewriting the OpGraph for optimization. It also has some header files to drive the development of OpGraph optimizations.


libquery/src/OpGraph.cpp
```diff
@@ -6,6 +6,34 @@
 
 namespace galois {
 
+void OpGraph::RemoveOp(Operator* op) {
+  if (op->NumChildren() > 1) {
+    GALOIS_LOG_FATAL("Does not support removing Ops with more than one child.");
```
roshandathathri, line 11: What's the reason for this assumption?

bowu: It complicates the removal while I don't see a need to support it for now. We can revisit it later.

PR #380, opened 2020-08-19
cusp/tsuba: support writing out distributed graphs

tylershunt:
Treat arrays needed by the partitioned graph like node and edge
properties so that we can store them. I added an option to the test macro
`STORE_PART` which adds a step to run a standalone partitioner before executing
the test.

~Not merge-able yet because cusp changed a lot while I was figuring out how best to add
tests, but thought I would put it up for comments.~

Now rebased to be compatible with HEAD


scripts/result_checker.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
ddn0, line 1: if this is going to be python3, I'd remove `from __future__ import print_function`, the uses of the `print` statement and the `if sys.version_info >= (3, 0):` checks

libcusp/include/galois/graphs/DistGraph.h
```diff
@@ -620,8 +624,145 @@ class DistGraph {
...
+
+    dist_graph->determineThreadRanges();
+    dist_graph->determineThreadRangesMaster();
+    dist_graph->determineThreadRangesWithEdges();
+    dist_graph->initializeSpecificRanges();
+    return std::unique_ptr<DistGraph>(std::move(dist_graph));
+  }
+
+  template <typename ArrowBuilderTy, typename Container>
+  static Result<std::shared_ptr<arrow::ChunkedArray>>
```
ddn0, line 691: I'm wondering if some basic functionality like this should go into `libgalois/include/.../Properties.h`

scripts/result_checker.py
tylershunt: I really just wanted `os.scandir` and I can work around that if people are resistant to moving to python3, but it also seems fine to go python3 since it's getting progressively unreasonable not to have it.

libcusp/include/galois/graphs/DistGraph.h
ddn0: In general, I'm thinking some basic `arrow::Array` to `T[]` or `uint8_t` and vice versa might be nice. I haven't had much fun using the typed interfaces on top of `arrow::Array` like `arrow::UInt32Array` because they kind of get in the way when you are working at the data serialization level (vs the application level).

scripts/result_checker.py
ddn0: If we are going with Ubuntu 18.04 as our oldest supported platform, it ships with Python 3: https://askubuntu.com/questions/1075203/python-version-in-ubuntu-18-04

libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -3173,7 +3176,36 @@ PartitionGraph(const DistOptions& opt) {
...
+    return std::make_unique<Impl>(net.ID, net.Num, opt);
+  }
+
+  // allow graph "partitioned" for a single host to be repartitioned if
+  // requested
+  if (stat.num_hosts == 1 && net.Num != 1) {
+    return std::make_unique<Impl>(net.ID, net.Num, opt);
+  }
+
+  if (stat.num_hosts != net.Num || stat.transpose != opt.transpose ||
```
gurbinder533, line 3198: NOTE: Once we have in-memory transpose, we can relax the "transpose" condition here. Users can choose to iterate in-edges or out-edges regardless of the partitioning policy used, which means the transpose flag should be recalculated based on policyID and edges to iterate.

libcusp/include/galois/graphs/DistGraph.h
tylershunt: Yeah it's pretty clunky; Not sure what's actionable for this PR, I'm happy to move this up but it seems like you want something more and I'm not quite sure what that is.

tylershunt:
Updated to respect changes to HEAD

scripts/result_checker.py
tylershunt: Sticking with python3 unless someone else complains.

libcusp/include/galois/ArrowIterator.h
```diff
@@ -12,6 +15,20 @@ std::shared_ptr<T> safe_cast(const std::shared_ptr<U>& r) noexcept {
   return p;
 }
 
+template <typename ArrowBuilderTy, typename Container>
+static Result<std::shared_ptr<arrow::ChunkedArray>>
+ArrayFromContainer(const Container& cont) {
```
ddn0, line 20: um... ArrowIterator is going away soon.

If you want just basic copy capability, I think https://github.com/apache/arrow/blob/master/cpp/src/arrow/stl.h might have something for you (though it mainly builds tables rather than columns). `TableFromTupleRange`

libcusp/include/galois/graphs/Distribution.h
```diff
@@ -208,6 +228,120 @@ class Distribution {
...
+
+    transposed_           = part_meta->transposed_;
+    is_vertex_cut_        = part_meta->is_vertex_cut_;
+    cartesian_grid_       = part_meta->cartesian_grid_;
+    num_global_nodes_     = part_meta->num_global_nodes_;
+    num_global_edges_     = part_meta->num_global_edges_;
+    num_nodes_            = part_meta->num_nodes_;
+    num_edges_            = part_meta->num_edges_;
+    num_owned_            = part_meta->num_owned_;
+    begin_master_         = part_meta->begin_master_;
```
ddn0, line 245: begin_master is no longer used.

libcusp/include/galois/ArrowIterator.h
ddn0: this is more FYI, I can clean this up when I remove ArrowIterator.h

libcusp/include/galois/graphs/Distribution.h
```diff
@@ -208,6 +228,118 @@ class Distribution {
...
+    part_meta->num_nodes_with_edges_ = num_nodes_with_edges_;
+
+    auto ltg_res =
+        ArrayFromContainer<arrow::UInt64Builder>(local_to_global_vector_);
+    if (!ltg_res) {
+      return ltg_res.error();
+    }
+    part_meta->local_to_global_vector_ = std::move(ltg_res.value());
+
+    GALOIS_LOG_ERROR("DIST GRAPH WRITE Mirror nodes: {}", mirror_nodes_.size());
```
ddn0, line 301: ???

lonestar/CMakeLists.txt
```diff
@@ -112,8 +112,8 @@ if(GALOIS_ENABLE_DIST)
     endif()
   endfunction()
 
-  function(add_test_dist_and_verify app input type part N np)
-    set(options GPU NOT_QUICK NO_VERIFY)
+  function(add_test_dist_and_verify app input type part N np input_uri)
```
ddn0, line 115: I wonder if we should move some of these positional arguments as one value or multi value arguments, e.g.,

```cmake
add_test_dist(app INPUT my_input INPUT_URI uri)
 ```
Forwarding arguments might be a little less error-prone as well.

I don't mind the state as it is, but it seems to be growing into a tangle.

Comment (tylershunt) [reply lonestar/CMakeLists.txt:115]: Yeah was aiming for as few changes as possible, I needed a handle on the `input_uri` so that I could transform it.

Comment (tylershunt) [reply lonestar/CMakeLists.txt:115]: > Yeah was aiming for as few changes as possible, I needed a handle on the `input_uri` so that I could transform it. I'll do some quick clean up as well.

File: libcusp/include/galois/graphs/Partitioner.h (hunk 3 above)
Comment (tylershunt) [reply libcusp/include/galois/graphs/Partitioner.h]: I'll make an issue so that we remember to revisit this when that lands.

Comment (tylershunt) [reply libcusp/include/galois/graphs/Partitioner.h]: Mentioned in issue #385

File: libcusp/include/galois/graphs/Distribution.h [hunk 8]
```diff
@@ -208,6 +228,117 @@ class Distribution {
...
+    ArrowIterator<arrow::CTypeTraits<uint32_t>::ArrayType> gtl_vals(
+        part_meta->global_to_local_values_.get());
+    auto k_it = gtl_keys.begin();
+    auto v_it = gtl_vals.begin();
+    for (; k_it != gtl_keys.end(); ++k_it, ++v_it) {
+      global_to_local_map_.emplace(*k_it, *v_it);
+    }
+
+    for (uint32_t i = 0; i < part_meta->mirror_nodes_.size(); ++i) {
+      mirror_nodes_[i].clear();
```
witchel, line 266: Number of mirror nodes always fits in a unit32_t?

libtsuba/include/tsuba/RDG.h
```diff
@@ -48,11 +48,19 @@ struct PropertyMetadata {
...
+  bool transpose;
+};
+
 // Struct version of main graph metadatafile
 struct GALOIS_EXPORT RDGMeta {
-  uint64_t version;
-  uint64_t previous_version;
-  uint32_t num_hosts;
+  uint64_t version{0};
+  uint64_t previous_version{0};
```
witchel, line 60: What is the deal with 0 literal initializing a uint64_t?  Will that work?  Is 0 a 32-bit constant?

libtsuba/src/RDG.cpp
```diff
@@ -43,24 +54,34 @@ namespace tsuba {
...
-  if (j.count("previous_version") != 0) {
-    j.at("previous_version").get_to(meta.previous_version);
-  } else {
-    meta.previous_version = 0UL;
-  }
   j.at("num_hosts").get_to(meta.num_hosts);
+
+  // these values are optional
+  if (auto it = j.find("previous_version"); it != j.end()) {
+    it->get_to(meta.previous_version);
```
witchel, line 76: Hmm, find does seem superior to count.

libtsuba/include/tsuba/RDG.h
```diff
@@ -48,11 +48,19 @@ struct PropertyMetadata {
   std::string path;
 };
 
+struct GALOIS_EXPORT RDGStat {
+  uint64_t num_hosts;
+  uint32_t policy_id;
```
witchel, line 53: {0} defaults?

libcusp/include/galois/graphs/Distribution.h
tylershunt: Same as the number of hosts.

libtsuba/include/tsuba/RDG.h
tylershunt: afaik the only time bit width of constants matters is when you're doing constant math where bit width matters like shifting.

libtsuba/include/tsuba/RDG.h
tylershunt: Sure I can add those.

libcusp/include/galois/graphs/Distribution.h
witchel: I should read up on this.  The mirror node isn't like a node in the graph, it is like a virtual host?

tylershunt: This is an array of arrays, each array in  `mirror_nodes_` corresponds to one host.

tylershunt:
I'll wait until @gurbinder533 signs off before I merge.

libcusp/include/galois/graphs/DistGraph.h
```diff
@@ -272,9 +276,26 @@ PartitionGraph(const DistOptions& opt) {
...
+    assert(stat.num_hosts == 1);
+    Partitioner<NodeData, EdgeData, Policy> part(dist.get(), opt);
+    part.Partition();
+  } else if (stat.num_hosts == net.Num && stat.transpose == opt.transpose &&
+             p_id == Policy::ID) {
+    if (auto res = dist->LoadFromStorage(opt.filename); !res) {
+      GALOIS_LOG_FATAL("Failed to populate distribution: {}", res.error());
+    }
+  } else {
+    GALOIS_LOG_FATAL("Graph requested is already partitioned");
```
gurbinder533, line 297: This error message is slightly confusing. This should fail if graph was partitioned but it doesn't match the required state such as hosts, transpose, etc., right? 

Shouldn' it just re-partition in this case?

libcusp/include/galois/graphs/Distribution.h
gurbinder533: mirror_node_ is an overloaded term here. Node here actually refers to a vertex proxy created in the graph but as Tyler said it is the vector of vectors (mirror_nodes_[host_id][node_id]).

libcusp/include/galois/graphs/PartitioningPolicies.h
```diff
@@ -11,8 +11,23 @@
...
+  kMiningPolicyNaiveID,
+  kMiningPolicyDegreesID,
+  kGenericCVCID,
+  kGenericCVCColumnFlipID,
+  kGenericHVCID,
+  kGingerPID,
+  kFennelPID,
+  kSugarPID,
+  kSugarColumnFlipPID,
+};
```
gurbinder533, line 26: These policy names have been updated, but I can take care of with my PR.

libcusp/include/galois/graphs/DistGraph.h
tylershunt: How do we re-partition? My understanding was that we have to start from an un-partitioned graph.

PR #395, opened 2020-08-21
Re-structure of Table to be better templated, and more encapsulated

danielmawhirter:
With the goal of only having a single switch over types for the Table, this PR templates out the type-dependent functionality and cleans up some of the semantics surrounding Table. Iteration to continue in the future.


libquery/src/Table.cpp
```diff
@@ -137,60 +169,79 @@ TableBuilder TableBuilder::BuildFrom(
...
-std::shared_ptr<arrow::Table> Table::Get() { return table_; }
-const std::shared_ptr<arrow::Table> Table::Get() const { return table_; }
+galois::Table::Table(std::shared_ptr<arrow::Table> table) : table_(table) {}
+template <typename T>
+std::optional<T> galois::Table::GetColumn(size_t index, size_t columnNumber) {
+  using ArrowType = typename arrow::CTypeTraits<T>::ArrowType;
+  using ArrayType = typename arrow::TypeTraits<ArrowType>::ArrayType;
+  assert(CreateDataType<T>::Call()->Equals(
+      table_->schema()->field(columnNumber)->type()));
+  std::shared_ptr<arrow::ChunkedArray> chunked = table_->column(columnNumber);
```
ddn0, line 203: - I'd like to centralize how we access Arrow arrays. Is there a reason we can't use `Properties.h` here (or improve it to suit)?
- BTW one of these functions `Table::column` or `ChunkedArray::chunk` (I forget which) actually copies the shared_ptr, which means you are going to incur an atomic increment every `GetColumn`.

danielmawhirter: This is related to a concern where I'm tempted to use arrow::Array, but stop using arrow::Table. Directly using arrow::Array should resolve point 2, since it bypasses most of Arrow's machinery for accesses. It also makes me more confident with adopting Properties.h as the view, though it might be worth a pass on naming conventions there just to make it more legible when it contains things like vertex id's. Overall I think I agree with these points, and would plan on implementing them soon, but not necessarily in this PR

libquery/src/Table.cpp
```diff
@@ -9,120 +8,152 @@ inline void ARROW_ASSERT(arrow::Status status) {
...
+  // using namespace table_detail;
   assert(column_number < schema_.NumColumns());
-  assert(arrow::Type::UINT32 == schema_.Type(column_number));
-  ARROW_ASSERT(
-      static_cast<arrow::UInt32Builder*>(builders_.at(column_number).get())
-          ->Append(value));
+  // assert(CreateDataType<T>::Call().id() == schema_.GetType(column_number));
+  using ArrowType   = typename arrow::CTypeTraits<T>::ArrowType;
+  using BuilderType = typename arrow::TypeTraits<ArrowType>::BuilderType;
+  ARROW_ASSERT(static_cast<BuilderType*>(builders_.at(column_number).get())
```
ddn0, line 110: all these `static_cast` s through `shared_ptr`s should be written with `std::static_pointer_cast` or `std::dynamic_pointer_cast` . The pointer casts take care of linking the derived pointer with the source pointer.

libquery/src/Table.cpp
```diff
@@ -137,60 +168,79 @@ TableBuilder TableBuilder::BuildFrom(
...
+template <typename T>
+std::optional<T> galois::Table::GetColumn(size_t index, size_t column_number) {
+  using ArrowType = typename arrow::CTypeTraits<T>::ArrowType;
+  using ArrayType = typename arrow::TypeTraits<ArrowType>::ArrayType;
+  assert(CreateDataType<T>::Call()->Equals(
+      table_->schema()->field(column_number)->type()));
+  std::shared_ptr<arrow::ChunkedArray> chunked = table_->column(column_number);
+  std::shared_ptr<arrow::Array> array          = chunked->chunk(0);
+  // TODO(danielmawhirter) bad assumption, unless constructed by TableBuilder...
+  assert(array->length() == chunked->length());
```
l-hoang, line 205: I think we're assuming that all chunked arrays only have 1 chunk for performance reasons.

libquery/src/Table.cpp
```diff
@@ -9,120 +8,152 @@ inline void ARROW_ASSERT(arrow::Status status) {
...
-    AddColumn(field->name(), field->type()->id());
+template <typename T>
+struct CreateDataType {
+  static std::shared_ptr<arrow::DataType> Call() {
+    using ArrowType = typename arrow::CTypeTraits<T>::ArrowType;
+    return arrow::TypeTraits<ArrowType>::type_singleton();
+  }
+};
+
+// This function requires at least c++14 support
```
l-hoang, line 40: This comment can be removed since we require C++17

libquery/include/querying/Table.h
```diff
@@ -8,50 +8,54 @@
...
+    std::string GetAlias(size_t index) const;
+    Type GetType(size_t index) const;
 
-  size_t NumColumns() const;
-  std::string Alias(size_t index) const;
-  arrow::Type::type Type(size_t index) const;
+  private:
+    std::vector<std::pair<std::string, Type>> schemaArray_;
+  };
+  class Builder {
```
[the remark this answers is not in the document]
danielmawhirter: It's mostly intentional, small note though, the exposed members of the galois namespace are Table, Table::Builder, and Table::Schema

PR #397, opened 2020-08-21
python: Support numpy dtype structs in Galois data and loops.

arthurp:
This adds struct support to InsertBag, LargeArray, and the loops do_all and for_each, and OBIM metrics. This also adds support for some additional primitive types (32-bit types).

To implement this, I refactored a lot of the wrapping utilities and automation. I tried to make the use of these more consistent as well so that the code is easier to understand.

There are also some other small commits making additional changes that I did along the way.

By using structs in work lists, SSSP is 3× faster than without structs. This is because structs in the work lists allow the algorithm to avoid duplicating work when the same node is added to the work list more than once (due to multiple paths to that node).


python/galois/sssp_property_graph.py
```diff
@@ -26,32 +30,34 @@ def create_distance_array(g: PropertyGraph, source):
 
 
 @for_each_operator()
-def sssp_operator(g: PropertyGraph, dists: np.ndarray, edge_lengths, nid, ctx: UserContext):
-    # TODO: We cannot distinguish multiple paths to the same node we we have to do the full amount of work each time.
-    #  With some kind of structures as iteration elements (instead of just nid) we could short cut the case where we
-    #  have already
-    for ii in g.edges(nid):
+def sssp_operator(g: PropertyGraph, dists: np.ndarray, edge_lengths, item, ctx: UserContext):
```
gurbinder533, line 33: edge_lenghts --> edge_weights

arthurp:
> Does it allow arbitrary structs or are there some restrictions in terms of the total size of data-types that can be in a struct?

This allows any numpy struct up to 128 bytes in total size. It will waste space for large structs since there are fewer implementations in that larger size ranges. The current set of supported sizes are: 8, 16, 32, 48, 64, 128. The system will pick the smallest size that fits the struct the user provides. We can pick whatever list of sizes we want, I just guessed, the only limitation is that a larger list makes for longer build times and larger binaries.

arthurp:
> Looks good. I didn't completely understand the jinja and Numba magic though.

Honestly, I'm happy if you even somewhat understood the Jinja and Numba magic. That code is messy and hard to understand. I'm worried it's moving towards being "write-only" (i.e., totally unreadable). Not that we have time, but I would love to spend some time to redesign it to be cleaner and easier to understand.Though honestly, I'm not sure it's possible to make it clear within the frameworks we are working in.

PR #399, opened 2020-08-21
FileView Read Ahead

scober:
A simple prefetching algorithm for `FileView`!

It gives about 10% improvement on read performance for parquet files that are large enough to span several row groups. I think 10% is about as good as we can do without a more sophisticated (i.e. parquet metadata aware) strategy. That is because (1) reading from S3 has high overhead and (2) we don't want to over-prefetch because all outstanding reads must be resolved when destroying the `FileView`.

It should have minimal effect on non-parquet uses of `FileView`.

I also snuck in some changes to fo_bench because I modified it a lot while I was trying to improve the prefetching strategy.


libtsuba/src/FileView.cpp
```diff
@@ -106,36 +112,50 @@ galois::Result<void> FileView::Bind(const std::string& filename, uint64_t begin,
...
+    return ErrorCode::InvalidArgument;
+  }
   // Gracefully handle the fill zero case here to simplify Bind
   if (in_end != in_begin) {
-    if (auto opt = must_fill(page_number(in_begin), page_number(in_end));
+    if (auto opt =
+            must_fill(filling_, page_number(in_begin), page_number(in_end));
         opt.has_value()) {
       std::tie(first_page, last_page) = opt.value();
       found_empty                     = true;
```
witchel, line 135: Can you use a structured binding here?  auto [first_page, last_page] = opt.value();

libtsuba/src/FileView.cpp
```diff
@@ -362,24 +410,69 @@ std::optional<std::pair<uint64_t, uint64_t>> FileView::must_fill(uint64_t begin,
...
 }
 
-galois::Result<void> FileView::mark_filled(uint64_t begin, uint64_t end) {
+galois::Result<void> FileView::mark_filled(uint64_t* bitmap, uint64_t begin,
+                                           uint64_t end) {
   uint64_t begin_mask;
   if (begin % 64) {
-    begin_mask = (1UL << (64 - begin % 64)) - 1UL;
+    begin_mask = (UINT64_C(1) << (64 - begin % 64)) - UINT64_C(1);
   } else {
```
witchel, line 418: Ah, UINT64_C(), finally, the one true way for constants.

libtsuba/src/FileView.cpp
```diff
@@ -362,24 +410,69 @@ std::optional<std::pair<uint64_t, uint64_t>> FileView::must_fill(uint64_t begin,
...
+  for (auto it = fetches_->begin(); it != fetches_->end();) {
+    auto fetch = it;
+    if (fetch->first_page <= page_number(start + size) ||
+        fetch->last_page >= page_number(start)) {
+      // Complete the remaining work if there is some
+      if (fetch->work != nullptr) {
+        while (!fetch->work->Done()) {
+          if (auto res = (*fetch->work)(); !res) {
+            return res.error();
+          }
```
witchel, line 452: In general our code for dealing with transient failures is lame, but if this loop's intent is to resolve outstanding prefetches and something goes wonky in the middle, it is best to log the error and tough it out or return early?  I think I'd go former, but I don't have solid context.

libtsuba/src/FileView.cpp
```diff
@@ -362,24 +410,69 @@ std::optional<std::pair<uint64_t, uint64_t>> FileView::must_fill(uint64_t begin,
...
+  }
+  return galois::ResultSuccess();
+}
+
+galois::Result<void> FileView::PreFetch(int64_t start, int64_t size) {
+  // Our highly sophisticated prefetching algorithm is to crudely approximate
+  // the size of the last read plus 10%. This is largely motivated by parquet
+  // files, which consecutively read row groups that are (in theory)
+  // approximately the same size.
+  int64_t fetch_size = (size / 10) * 11;
```
witchel, line 468: As the man says in the song, the crux of the biscuit is the apostrophe.

If your prefetch policy is to read ahead 10% and you get a 10% performance gain, one might argue that your heuristic is insufficiently aggressive.

I don't think this issue is worth ratholing, but I would try a more aggressive policy, like read twice as much as requested.  Because we know the library that is making the calls, we should be able to tailor our prefetching strategy. 

I'm having trouble following the logic in this checkin, but our prefetches are issued as async reads?  That should be ok.  If you are concerned about priority inversion (prefetch reads being serviced before demand reads), you can async read the prefetch and sync read the demand read.

But again, if I were a brain in a vat, you could get me to talk prefetch strategies for days without eating or sleeping.

libtsuba/include/tsuba/FileView.h
```diff
@@ -122,17 +146,27 @@ class GALOIS_EXPORT FileView : public arrow::io::RandomAccessFile {
...
+  inline uint64_t f_page(uint64_t* bitmap, uint64_t block_num, uint64_t start,
+                         uint64_t end);
 
   // Given a starting and ending page, return an inclusive region that describes
   // which pages must be fetched from storage. Or an empty std::optional if no
   // pages need to be fetched.
-  std::optional<std::pair<uint64_t, uint64_t>> must_fill(uint64_t begin,
-                                                         uint64_t end);
+  std::optional<std::pair<uint64_t, uint64_t>>
+  must_fill(uint64_t* bitmap, uint64_t begin, uint64_t end);
```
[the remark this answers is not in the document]
scober: Makes sense

libtsuba/src/FileView.cpp
```diff
@@ -94,10 +99,11 @@ galois::Result<void> FileView::Bind(const std::string& filename, uint64_t begin,
 
   map_start_ = static_cast<uint8_t*>(tmp);
   mem_start_ = -1;
-  present_ =
+  filling_ =
```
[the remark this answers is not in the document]
scober: I agree, and I think it may also make sense to change the signatures of all the helper functions to not expect `uint64_t*`.

libtsuba/src/FileView.cpp
```diff
@@ -106,36 +112,50 @@ galois::Result<void> FileView::Bind(const std::string& filename, uint64_t begin,
...
-    if (auto opt = must_fill(page_number(in_begin), page_number(in_end));
+    if (auto opt =
+            must_fill(filling_, page_number(in_begin), page_number(in_end));
         opt.has_value()) {
       std::tie(first_page, last_page) = opt.value();
       found_empty                     = true;
     }
 
     uint64_t file_off = first_page * (1UL << page_shift_);
-    ;
```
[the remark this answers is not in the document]
scober: She sacrificed herself for the good of us all. Never forget.

libtsuba/src/FileView.cpp
scober: I forget why I used `std::tie` instead of `auto [first, last]`. It may have had something to do with the return type of `opt.value()`? Is there a big difference between the two?

libtsuba/src/FileView.cpp
scober: I'm not sure. My feeling was that if the outstanding work returned an error, that meant that something had gone horribly wrong and attempting to continue would not create useful results. But I guess it is just an interface question? Do we want `AsyncWork` to return errors even when it thinks it can continue? And if so how should it signal that it cannot continue?

libtsuba/src/FileView.cpp
scober: I wish I had thought about this this way. I have already ratholed. I tried a wide variety of prefetching policies (bigger, smaller, multiple pieces) and they all gave about a 10% performance gain. I think the issue is that S3 reads have high overhead and as long as you are reading in chunks that are on the order of 1GB, you will incur high overhead costs. That 10% improvement seems to be a fundamental tension between the conversion throughput and the S3 read throughput (when reads are relatively small). I actually think I can break through that barrier, but I would like to land these changes first.

And yes, the basic logic of these changes is "I think you will need these bytes next, so let me start reading them (asynchronously) now". And if a read comes in that is not covered by an outstanding async read I fetch it from S3 synchronously before I do anything else. Although I am not sure that that actually resolves the priority inversion problem. I think the bottleneck that would require prioritization is if the network or the tsuba thread pool became saturated. And I don't think that synchronous reads get priority when there is contention there.

libtsuba/include/tsuba/FileView.h
```diff
@@ -9,9 +9,18 @@
 #include "galois/config.h"
 #include "galois/Result.h"
 #include "galois/Logging.h"
+#include "tsuba/FileAsyncWork.h"
 
 namespace tsuba {
 
+namespace internal {
+struct FillingRange {
```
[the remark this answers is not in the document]
scober: I don't understand this comment.

If I change this line to `class FillingRange;`, can I declare the class members separately? Somewhere else?

What is the issue with the  FileView destructor?

And should this not be internal? I thought that marking this internal would be a good hint that this is not intended to be a generally useful class.

libtsuba/src/FileView.cpp
ddn0: I suggested `std::tie` for here because `first_page` and `last_page` were already declared as variables earlier. If you want to declare and unpack in one statement, `auto [x, y] = ...` is the way to go.

scober: And I think I needed to do it that way because `first_page` and `last_page` need to outlive the scope of this if block.

libtsuba/include/tsuba/FileView.h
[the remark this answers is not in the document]
ddn0: I was just referring to this idea: https://en.cppreference.com/w/cpp/language/pimpl and if you follow this technique, `unique_ptr` needs to be able to refer to the destructor, which is why I was referring to `FileView`.

I suppose now that we have `-fvisibility=hidden` by default this matters less. But I normally think of `internal` namespaces as a hack when you cannot find any other way to keep internal details private (e.g., you can't use statically linked functions, private methods). Internal namespaces are just a convention and may bleed into an API anyway if you aren't careful. 

In this particular case, you might be able to hide the type more strongly using pimpl or perhaps more simply by just keeping `FillingRange` as `tsuba::FillingRange` but not `GALOIS_EXPORT`ing it.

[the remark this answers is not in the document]
scober: I did not change it to a pimpl implementation, but I did change `tsuba::internal::FillingRange` to `tsuba::FileView::FillingRange` which I agree hides it better.

PR #405, opened 2020-08-24
Simplifying the code by removing unused DFS template option. This opt…

roshandathathri:
…ion will not be valid once we move to distributed execution.


ddn0:
nit about the commit message: it would be good to keep the length of the first line to below 72 characters otherwise you get this very odd wrapping in most tools: https://github.com/KatanaGraph/katana/pull/405/commits

Generally, you'd write:
```
Imperative voice

Some details written in paragraphs. These are hard wrapped to 72 characters.
```

E.g.,

```
Remove unused DFS template option

This option will not be supported during distribution execution so prune it away now.
```

PR #406, opened 2020-08-24
Initial NOT support for expression trees

l-hoang:
Adds NOT support for expression trees by adding the NOT operator that
will only use the left child of an expression tree and negate whatever
truth value that child takes. Added the constructors and logic necessary
to handle this in the evaluate function for expression trees. Added a
set of tests to test the NOT parsing for expected behavior.

One thing to note is the behavior of NOT (pattern path). In the current
system, instead of creating an expression tree node for these paths, the
path is added directly to the query graph as an edge that should not
exist. In order to allow this behavior without breaking evaluation of the
expression tree, it is necessary to insert a dummy node for the pattern
path's parse. At the moment, it always returns true when evaluated, but
this is *not* correct behavior in all cases as you may need to know
when that pattern path condition is false. This will need to be
revisited in the future (the match of the pattern path should be tied to
an expression tree node).


libquery/src/CypherCompiler.cpp
```diff
@@ -765,19 +763,46 @@ void CypherCompiler::CompileLabelsOperator(CypherASTNode ast,
...
+      // make exp tree child of a not op
+      std::unique_ptr<ExpressionTree> not_exp_tree =
+          std::make_unique<ExpressionTree>(kTreeNot, std::move(exp_tree));
+      // return the variant
+      return CompilerTreeVariant(std::in_place_index<kExpressionTree>,
+                                 std::move(not_exp_tree));
     }
   }
+
+  GALOIS_LOG_WARN("An unsupported unary operation has been detected. Ignoring "
```
roshandathathri, line 802: Better to abort in my opinion but it's fine either way.

l-hoang: still trying to walk that fine line between "is this recoverable" or "meh just kill the program" since we don't want users to see an ugly failure message :P

PR #434, opened 2020-08-27
tsuba: Benchmark transactions

witchel:
Code to benchmark transactions.  150ms/tx for an EC2 machine.  I think we could easily get that under 100ms, but there are higher priorities.


tylershunt:
> The Google style guide says something like "use type inference when it will improve clarity", which is not super enlightening

CC @ddn0 

We don't have any hard or fast rules. I personally prefer to see the actual types in most cases. I think it improves clarity for cases where the type is already on the line (`make_unique`), your dealing with some status type that's used everywhere anyway and who's type will become clear quickly (`galois::Result`), or when your writing templated code that `auto` can simplify. The only other example that I can think of is like `iterators` and similar where you don't care what the type is.

In general I think it's just a judgment call.

ddn0:
> We don't have any hard or fast rules. I personally prefer to see the actual types in most cases. I think it improves clarity for cases where the type is already on the line (make_unique), your dealing with some status type that's used everywhere anyway and who's type will become clear quickly (galois::Result), or when your writing templated code that auto can simplify. The only other example that I can think of is like iterators and similar where you don't care what the type is.
>
> In general I think it's just a judgment call.

If you are looking for clear guidelines, what Tyler has written above, I think, strikes the right balance.

Explicit types are preferred. Use `auto` to avoid excessive duplication, e.g., use `auto` to avoid type stuttering within the same neighborhood (`make_unique`, `galois::Result`) or where the type itself does not matter (e.g., some parts of templated code).

Under these guidelines, `auto` for range iteration (i.e., `for (auto x : y)`) is probably not advised though it is pretty common in the codebase at the moment. Using `auto` for legacy-range iteration (i.e., `for (auto begin = v.begin(), end = v.end(); begin != end; ++begin) { MyType& t = *begin; .... }`) is a fine and good use of `auto`.

PR #440, opened 2020-08-28
Let movability flow from members

ddn0:
Also known as the rule of zero. If a class doesn't add any restrictions
on top of its members, it can derive its own copyability and movability
implicitly from that of its members.

A similar idea holds for containers of only movable elements like
vector<unique_ptr>. There is generally no need to pass such containers
by reference or pointer. Passing them by value gives the desired
semantics because callers must explicitly use std::move, which allows
callers to clearly express the intent that they no longer want (or can
have) the object.

Along the way:
- reduce the number of extra copies incurred by shared_ptrs. Each copy
  is an atomic increment. Prefer passing by reference or adding moves
  where possible.
- remove superfluous `this->` prefix to access members of a class. Class
  methods generally can access their members without any qualification.


libquery/include/querying/ops/FilterOp.h
```diff
@@ -6,12 +6,13 @@ namespace galois {
 
 class FilterOp final : public Operator {
 public:
-  FilterOp(std::vector<std::unique_ptr<ExpressionTree>>* expression_trees) {
+  FilterOp(std::vector<std::unique_ptr<ExpressionTree>> expression_trees) {
     // move each tree to this op node's vector of trees
-    for (auto&& ptr : *expression_trees) {
-      expression_trees_.emplace_back(std::move(ptr));
+    auto tmp = std::move(expression_trees);
```
l-hoang, line 11: does this move leave the original object in a clearable state?

libquery/include/querying/ExpressionTree.h
```diff
@@ -94,13 +94,6 @@ enum ExpressionTreeOperator {
 //! TODO(l-hoang) probably makes more sense to call it a literal tree
 class ArithmeticTree {
 public:
-  ArithmeticTree()                      = delete;
```
l-hoang, line 97: is the default constructor delete implicit with the definition of the other constructors?

libgalois/include/galois/graphs/LC_CSR_Graph.h
```diff
@@ -55,7 +55,6 @@ template <typename NodeTy, typename EdgeTy, bool HasNoLockable = false,
           typename FileEdgeTy = EdgeTy>
 class LC_CSR_Graph :
     //! [doxygennuma]
-    private boost::noncopyable,
```
l-hoang, line 58: what was the point of this existing in the first place?

libquery/include/querying/ops/FilterOp.h
ddn0: ?

Even before the move, you can consider the original object from the caller as empty as `expression_trees` has already been moved constructed to make this call.

I suppose this move is somewhat superfluous. I could have written:
```cpp
auto& trees = expression_tress;
....
```
I didn't think of it at the time because taking references of parameters is not normally something I consider. Is that what you are asking about?

libquery/include/querying/ExpressionTree.h
ddn0: The language of the spec is along the lines of the default constructor is implicitly defined if a class does not have a constructor. The implicitly defined default constructor is then implicitly deleted if the class has members that would make the implicitly defined constructor invalid (e.g., members are default constructable themselves).

https://en.cppreference.com/w/cpp/language/default_constructor

https://godbolt.org/z/MEEqxe

Long story short: if a class defines a constructor it will not have a default constructor unless you explicitly add one.

libgalois/include/galois/graphs/LC_CSR_Graph.h
ddn0: Before C++11, the common way to make a class noncopyable was to declare the copy functions private and but not implement them. `boost::noncopyable` is a simple utility class that does this for you.

This doesn't play very well with how special member functions are defaulted and deleted post C++11. In particular, it was preventing subclasses or classes with LC_CSR_Graph members (I forget which) from being considered movable implicitly.

libquery/include/querying/ops/FilterOp.h
l-hoang: i guess what i'm asking is if the original object is still "valid" in the sense that i can still call its class members and such (e.g. vector.clear()) since that is what occurs later in the code

libquery/include/querying/ExpressionTree.h
l-hoang: sweet, thanks

libgalois/include/galois/graphs/LC_CSR_Graph.h
l-hoang: i'll never understand all of C++ i guess :P

libquery/include/querying/ops/FilterOp.h
ddn0: I don't see the `vector.clear()` you are referring to so I can't speak about the specifics, but in general, after a move construction, the original object is some unspecified but valid state. So, if `v.clear()` is valid to call in any vector state, it would fine to call `v.clear()` after a move. Stylistically, though, we'd prefer avoid mentioning the post-move object at all.

PR #441, opened 2020-08-28
Add support to cypher-query node properties

roshandathathri:
This commit only supports checking equality for node properties of type integer and string. It assumes integer to be uint32_t.


libquery/include/querying/SubgraphQuery.h
```diff
@@ -1,28 +1,20 @@
...
 //! Get subgraph matches from data graph given some query graph.
 //!
 //! Matches are returned in the returned_embeddings variable (bag containing the
 //! nodes comprising a match). The matching order used is also returned to the
 //! caller for post processing.
-size_t SubgraphQuery(QueryGraph& query_graph, DataGraph& data_graph,
+size_t SubgraphQuery(QueryGraph& query_graph,
+                     const QueryGNodeProperties& query_node_props,
+                     const QueryGEdgeProperties& query_edge_props,
+                     DataGraph& data_graph,
```
l-hoang, line 16: the attributed graph contains the data graph, so passing in both seems superfluous unless you don't want to expose the data graph from the attributed graph

also, this function has way too many arguments now

libquery/include/galois/graphs/QueryGraph.h
```diff
@@ -19,12 +19,40 @@
...
 
+// TODO needs to be generalized; also might be useful to do something similar
+// to dist apps where you can use a tag to get the index of a particular type
+//! A property value from a map can only be a literal at this point in time:
+//! string, int, or float Note that the 4th template argument is also a string;
+//! it's separated from the first to distinguish between a regular string and a
+//! string that's a parameter
+// using PropertyValue = std::variant<std::string, int, float, std::string>;
+
+// TODO handle float properties and parameters: see above unused comment
```
l-hoang, line 42: TODO(roshan), most likely

also this change means you lose the distinction between parameter name to string for the time being (which is noted by your PR saying this only works with ints), so it would likely help you all to place a TODO for the parameters as well

libquery/src/SubgraphQuery.cpp
```diff
@@ -32,8 +32,16 @@ class SubgraphQueryMiner {
 
   //! graph to do matching on
   DataGraph& dataGraph;
+  //! graph that contains the attributes for the data graph
+  // TODO(Sam) encapsulate this better
+  graphs::AttributedGraph& attributed_graph_;
```
l-hoang, line 37: const if they aren't altered

roshandathathri:
> you may want to leave a TODO somewhere saying that edge attribute matching needs to be done as well

The TODO is there is SubgraphQuery.cpp

libquery/include/querying/SubgraphQuery.h
roshandathathri: Yes, I don't want to expose data graph in attributed graph and yes, these need to be encapsulated better. There's a TODO in SubgraphQueryMiner for that.

libquery/include/galois/graphs/QueryGraph.h
roshandathathri: Hehe. Yes, I didn't assign that for now.

What is parameter name doing? I thought it was for runtime parameters?

libquery/src/SubgraphQuery.cpp
roshandathathri: Will do this after #440 is merged.

libquery/include/galois/graphs/QueryGraph.h
l-hoang: yes, runtime parameters

roshandathathri: That's not supported for now. So, yes, it's a TODO

libquery/include/galois/graphs/QueryGraph.h
```diff
@@ -19,12 +19,53 @@
...
+//! it's separated from the first to distinguish between a regular string and a
+//! string that's a parameter
+// using PropertyValue = std::variant<std::string, int, float, std::string>;
+
+// TODO handle parameters: see above unused comment
+using ExpressionValueType = std::variant<bool, std::string, uint32_t, int32_t,
+                                         uint64_t, int64_t, float>;
+//! Enum to make indexing into the ExpressionValueType more sane/readable
+//! Should match the order in ExpressionValueType
+enum ExpressionValueIndex {
```
ddn0, line 47: In general, I'm not a huge fan of using indices in addition to types because it just seems like a recipe for inconsistency. https://en.cppreference.com/w/cpp/utility/variant/visit seems like the expected way to dispatch on variants.

Regardless of the above, why was ExpressionValueType moved to this file? I'd expect the definition of ExpressionValues to be associated with the ExpressionTree and not the QueryGraph.

libquery/include/galois/graphs/QueryGraph.h
```diff
@@ -85,6 +126,10 @@ using QueryGraph = galois::graphs::LC_CSR_CSC_Graph<QueryNode, QueryEdgeData,
                                                     false, true, true>;
 //! QueryGraph node typedef
 using QueryGNode = QueryGraph::GraphNode;
+//! QueryGraph node properties
+using QueryGNodeProperties = std::vector<PropertyMap>;
```
ddn0, line 130: `QueryNodeProperty`, `QueryEdgeProperty`? I don't really know what an extra `G` adds.

libquery/include/galois/graphs/QueryGraph.h
```diff
@@ -19,12 +19,53 @@
...
+// TODO needs to be generalized; also might be useful to do something similar
+// to dist apps where you can use a tag to get the index of a particular type
+//! A property value from a map can only be a literal at this point in time:
+//! string, int, or float Note that the 4th template argument is also a string;
+//! it's separated from the first to distinguish between a regular string and a
+//! string that's a parameter
+// using PropertyValue = std::variant<std::string, int, float, std::string>;
+
+// TODO handle parameters: see above unused comment
+using ExpressionValueType = std::variant<bool, std::string, uint32_t, int32_t,
```
ddn0, line 43: `ExpressionValue` is probably a better name of this. An `ExpressionValue` contains a value of an expression.

libquery/include/galois/graphs/QueryGraph.h
```diff
@@ -19,12 +19,53 @@
...
+//! Map from a string (property name) to a value, which can be a string, int, or
+//! float
+using PropertyMap = std::unordered_map<std::string, ExpressionValueType>;
+
+using VertexId           = uint32_t;
+using QueryEmbeddingType = std::vector<VertexId>;
+#ifdef GALOIS_ENABLE_DIST
+using QueryEmbeddingListType = DistEmbeddingList<QueryEmbeddingType>;
+#else
+using QueryEmbeddingListType = EmbeddingList<QueryEmbeddingType>;
```
ddn0, line 66: I'd drop `Type` here too for the same reason

libquery/src/SubgraphQuery.cpp
```diff
@@ -56,9 +64,23 @@ class SubgraphQueryMiner {
 
   //! returns true if node is to be dropped from consideration for the
   //! given query node
-  inline bool pruneNode(const QueryGNode& queryNodeID,
-                        const DataNode& dataNode) {
-    return !matchNodeLabel(queryGraph.getData(queryNodeID), dataNode);
+  inline bool pruneNode(const QueryGNode& query_node_id,
+                        const DataGNode& data_node_id) {
+    if (matchNodeLabel(queryGraph.getData(query_node_id),
```
ddn0, line 69: style nit: I'd return early to reduce the overall nesting depth of this function, e.g.,

```cpp
if (!matchNodeLabel(...)) {
  return true;
}
...
```

Fewer things to keep in your mind.

libquery/src/SubgraphQuery.cpp
```diff
@@ -56,9 +64,23 @@ class SubgraphQueryMiner {
...
+                       dataGraph.getData(data_node_id))) {
+      const PropertyMap& query_node_props = query_node_props_[query_node_id];
+      if (!query_node_props.empty()) {
+        for (auto& it : query_node_props) {
```
ddn0, line 73: `const auto&`?

libquery/src/SubgraphQuery.cpp
```diff
@@ -56,9 +64,23 @@ class SubgraphQueryMiner {
...
+          std::optional<ExpressionValueType> data_prop =
+              attributed_graph_.GetNodeAttribute(data_node_id, it.first);
+          if (!data_prop.has_value() || (it.second != data_prop.value())) {
```
ddn0, line 76: Normally, I'm not a big fan of operator overloading like this, but `std::optional` has overrides so it looks like a pointer, which allows you to use it very naturally:
```cpp
if (!data_prop || it.second != *data_prop) {

}
```

libquery/include/galois/graphs/QueryGraph.h
roshandathathri: This requires some clean-up. AttributedGraph return type for a property value is ExpressionValueType. I did not want to change that now because it affects table and filter related computation.

I can do this refactoring later, unless you think it'll be more painful to use/refactor down the line.

libquery/include/galois/graphs/QueryGraph.h
roshandathathri: Will do.

libquery/include/galois/graphs/QueryGraph.h
roshandathathri: Will do.

libquery/include/galois/graphs/QueryGraph.h
roshandathathri: Will do

libquery/src/SubgraphQuery.cpp
roshandathathri: Makes sense.

libquery/src/SubgraphQuery.cpp
roshandathathri: I was surprised that this compiled because query_node_props_ is a const reference. Apparently, auto deduces "const" as well:
https://stackoverflow.com/questions/52204122/difference-between-const-auto-and-auto-if-object-of-reference-is-const

Anyway, I'll change it const auto for better readability.

libquery/src/SubgraphQuery.cpp
roshandathathri: Thanks.

libquery/include/galois/graphs/QueryGraph.h
ddn0: If there is a cyclic dependency between tables, attributed graph and their evaluation, I would just pull out these definitions to something like `Expression.h`. Looking at the definitions, it seems a little odd that you can't just `#include ExpressionTree.h` in `QueryGraph.h` since there is no particular reason the former should depend on the latter.

roshandathathri: Yes, it makes more sense to include ExpressionTree.h in QueryGraph.h (we will need that for the future as well). That works, so I'll push that change.

PR #452, opened 2020-08-30
Fix querying of int64 attributes

roshandathathri:


libquery/src/CypherCompiler.cpp
```diff
@@ -213,11 +213,11 @@ PropertyMap GetPropertiesFromMap(CypherASTNode map_ast) {
...
       property_value.emplace<kEVString>(cypher_ast_string_get_value(ast_value));
     } else if (value_type == CYPHER_AST_INTEGER) {
-      // assumes integer is uint32_t
+      // assumes integer is int64_t
       // TODO get the property type from the data graph and
       // use the appropriate integer type
-      property_value.emplace<kEVUInt32>(
-          std::stoi(cypher_ast_integer_get_valuestr(ast_value)));
+      property_value.emplace<kEVInt64>(
+          std::stol(cypher_ast_integer_get_valuestr(ast_value)));
```
l-hoang, line 220: there's other instances of this (stoi vs stol) in the code that should be changed as well if you want to make sure this doesn't occur again (i'm assuming this was one of the problems?)

libquery/test/cypher-compile-test.cpp
```diff
@@ -739,7 +739,7 @@ int main() {
   //   AssertBasicReturn(cc);
```
l-hoang, line 739: when were these tests commented out? shouldn't they work?

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
```diff
@@ -0,0 +1 @@
+MATCH ({id:21})-[a:LIKES]->() RETURN count(a);
```
l-hoang, line 1: may you add another test that tests something that goes beyond the size of an int32? that's the id that will be used for the august deadline demo

libquery/src/AttributedGraph.cpp
```diff
@@ -203,6 +203,12 @@ AttributedGraph::GetNodeAttribute(uint32_t vid,
     if (res) {
       return res.value();
     }
+  } else if (dtype->Equals(arrow::int64())) {
+    std::optional<int64_t> res =
+        attribute_array.getNumericElement<arrow::Int64Type>((uint64_t)vid);
```
l-hoang, line 208: use a `static_cast`

libquery/src/CypherCompiler.cpp
roshandathathri: The main problem was GetNodeAttribute was not supporting properties of type int64. stol is needed instead of stoi only if the type is int64 instead of int32. This is not the case for other uses of stoi.

libquery/test/cypher-compile-test.cpp
roshandathathri: No, parameters are not parsed anymore.

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
roshandathathri: I tried 17592186053195 but there is no person with that id in the test dataset.

libquery/src/AttributedGraph.cpp
roshandathathri: That was code copied from above. A cast is not required. I dropped (uint64_t).

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
roshandathathri: I replaced 21 with 17592186053195 and set the expected result as 0.

l-hoang: Use 17592186044418; name should be Abdul Harris

libquery/src/CypherCompiler.cpp
l-hoang: there is a case of stoi for parsing the number in the code somewhere that was causing problems iirc; see the function that parses integers

l-hoang:
also, i guess @bowu will be looking into the corresponding "WHERE" version of this test query? if not i can spend a bit of time on it on monday

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
roshandathathri: The output is still 0. Are you sure that the above query returns a non-zero count for this?

libquery/src/CypherCompiler.cpp
roshandathathri: You mean GetASTInteger()? That's using int types. Anyway, that's not related to this fix.

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
l-hoang: oh, i was just testing for existence; wasn't testing for LIKES

use this instead

id: 4398046511147

count should be 15

roshandathathri: I added a different query with that id which returns a non-zero result.

libquery/src/CypherCompiler.cpp
l-hoang: yes; it's probably related to the WHERE version of this though (cc @bowu)

lonestar/querying/cpu/graph-query/cypher-queries/single_edge/likesid.q
roshandathathri: Done.

l-hoang: thanks!

libquery/src/CypherCompiler.cpp
roshandathathri: No, it isn't. GetASTInteger() is only used to parse skip and limit.

The issue with the WHERE bug is in the types supported by ArithmetricTree/ExpressionTree.

l-hoang: i see, thanks for pointing that out

PR #454, opened 2020-08-30
cmake: Move generated pyx/pxd files to binary build directory.

arthurp:
To make this work, I add the build directory to the include paths. This requires a further hack to the Cython CMake functions.


ddn0:
> This requires a further hack to the Cython CMake functions.

It looks less hacky to me from the perspective of a user of the Cython CMake functions. Is "hack" here more like in the sense of "additional modification of vendored Cython CMake functions"?

arthurp:
Yes. The "hack" is just modifying the UseCython.cmake file further. It is really cleaner. My only concern is that there might be corner cases where my changes to the Cython cmake files break things.

PR #461, opened 2020-09-01
Distributed graph querying

roshandathathri:
This PR enables distributing the input graph
for running cypher queries.

To do so, this PR includes new features in partitioning: 
- The property graph is partitioned/distributed, i.e., all node and edge properties are distributed. Currently, only POD, boolean, and string types are supported for properties.
- Mirror edges are created during partitioning. The policy for creating mirror edges depends on the partitioning policy but it is not customizable yet. The mirror edge is created on the host that would be the master for the reverse direction edge (if it existed). The mirror edge is created on a host only if that host does not have the master edge.

The attributed graph used for querying is expected to be
pre-loaded into the query server. The attributed graph
requries the DistGraph to be constructed. The DistGraph
itself is destroyed after attributed graph is constructed.

The attributed graph stores distribution information as it
provides a partitioned view of the graph. All node identifiers
in its interface (both input and output) are global node IDs.

The graph-query app has been moved to distributed
sub-folder because distributed execution support is on
by default.


libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -1417,39 +1467,21 @@ class Partitioner {
...
-              }
-
-              while (indicators[owner] != 2) {
-                ; // until initialized, loop
+              uint32_t mirror = GetEdgeMirror(src, dst, num_local_edges);
+              if (mirror == owner) {
+                continue;
               }
-              has_incoming_edge[owner].set(dst);
+              ComputeInspectionData(mirror, src, dst, global_offset,
```
l-hoang, line 1481: so this is pretty much adding another edge to the metadata here, correct?

libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -2350,33 +2412,21 @@ class Partitioner {
...
-                if (only_dest) {
-                  to_send[owner].emplace_back(gdst);
-                } else {
-                  to_send[owner].emplace_back(source[pview.GetEdgeOffset(*ee)]);
-                }
+              uint32_t mirror = GetEdgeMirror(src, dst, num_local_edges);
+              if (mirror == owner) {
+                continue;
               }
+              PrepareSendData<only_dest>(mirror, src, dst, cur_edge,
```
l-hoang, line 2428: and this is the actual send

libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -2350,33 +2412,21 @@ class Partitioner {
...
+              uint64_t edge_offset = pview.GetEdgeOffset(*ee);
               uint32_t owner =
-                  graph_partitioner_->GetEdgeOwner(src, gdst, num_local_edges);
-
-              if (opt_.edge_state_rounds > 1) {
-                host_loads_[owner] += 1;
+                  graph_partitioner_->GetEdgeOwner(src, dst, num_local_edges);
+              PrepareSendData<only_dest>(owner, src, dst, cur_edge, edge_offset,
+                                         source, sink, to_send);
+              if (!MirrorEdges) {
```
l-hoang, line 2421: would prefer if all mirror edge related logic was in its own block rather than using continue here; continue can sometimes make things hard to understand, and for me this is one of those cases (ditto for the metadata save for this)

also a comment explaining what exactly this MirrorEdge variable/block does would help

l-hoang:
also, having a test for this somehow to check correctness would be nice + i would rather you get another approval besides mine before you merge this

libcusp/include/galois/graphs/Partitioner.h
roshandathathri: Yes.

libcusp/include/galois/graphs/Partitioner.h
roshandathathri: Yes.

libcusp/include/galois/graphs/Partitioner.h
roshandathathri: Disagree. I prefer this style for both return and continue because it reduces nesting, which in my opinion leads to better readability. 

It's just a couple of lines for the mirror edges. I'll just add a comment saying something like "do the same as the owner"

libcusp/include/galois/graphs/Distribution.h
```diff
@@ -66,14 +66,14 @@ struct DistOptions {
 template <typename NodeTy, typename EdgeDataTy>
 class DistGraph;
 
-template <typename Policy>
+template <typename Policy, bool mirror_edges_ = false>
```
ddn0, line 69: probably just `template <typename Policy, bool mirror_edges>`?

I.e., no default value and no trailing underscore

libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -1372,6 +1372,56 @@ class Partitioner {
...
+    bool host_is_master_of_dest =
+        (owner == graph_partitioner_->RetrieveMaster(dst));
+
+    if (host_is_master_of_dest) {
+      return;
+    }
+
+    // A mirror must be created for destination node on that host
+    // since it will not be created otherwise
+    auto& bitset_status = indicators[owner];
```
ddn0, line 1406: `CopyableAtomic<char>&`?

libcusp/include/galois/graphs/Partitioner.h
```diff
@@ -1372,6 +1372,56 @@ class Partitioner {
...
+
+    // initialize the bitset if necessary
+    if (bitset_status == 0) {
+      char expected = 0;
+      bool result   = bitset_status.compare_exchange_strong(expected, 1);
```
ddn0, line 1411: probably would be nice to define some constants for 1, 2 and 0.

libcusp/include/galois/graphs/Distribution.h
```diff
@@ -66,14 +66,14 @@ struct DistOptions {
...
 class Partitioner;
 
 class Distribution {
   // TODO(ddn): remove friend classes in favor of defined API between
   // Distribution and Partitioners
 
-  template <typename Policy>
+  template <typename Policy, bool mirror_edges_>
```
ddn0, line 76: I'd prefer to keep the number of template parameters small. It pushes implementation details to users that often can be handled internally to the class.

- Is mirroring edges really independent of the partitioning policy?
- Even if it is, can we just take a boolean value and then just dispatch on it internal to the Partitioner class?

roshandathathri:
@all, existing tests currently for the these changes are negative tests in the sense that these changes should NOT be breaking existing tests. The real/positive tests for these changes will be once I make graph_query read partitioned graphs (which will also be in this PR).

libcusp/include/galois/graphs/Distribution.h
roshandathathri: Mirroring edges is currently independent of the partitioning policy. There's a specific way to mirror which is required for pattern matching. Nothing else is currently supported.

I'll try to modify the code to take a boolean value in options and do an internal template dispatch.

libcusp/include/galois/graphs/Partitioner.h
roshandathathri: I'll omit this for this PR

roshandathathri:
@reviewers,
- All existing tests for graph-query except one. The failing query (personidcompare.q) has a where clause comparing properties of two different nodes. This has been moved to unsupported list.
- There are lots of changes in this PR, so it can become overwhelming. I would suggest reviewing commit-by-commit. Each commit is self-contained, except that the tests come together only in the last commit.

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -117,23 +127,30 @@ class GALOIS_EXPORT AttributedGraph {
...
   //! FindEdge searches for the edge between src and dst.
   std::optional<DataGraph::edge_iterator> FindEdge(uint32_t src,
                                                    uint32_t dst) const {
-    return graph.findEdge(src, dst);
+    auto local_src = dist_->getLID(src);
+    auto local_dst = dist_->getLID(dst);
+    return graph.findEdge(local_src, local_dst);
   }
 
   edges_iterator edges(uint32_t src, const QueryLabel& label) const {
```
l-hoang, line 147: would leave a comment mentioning this should be a GID (same for all functions that convert it over to a lid)

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -117,23 +127,30 @@ class GALOIS_EXPORT AttributedGraph {
...
   }
 
-  size_t GetDegree(uint32_t node) const { return graph.getDegree(node); }
+  size_t GetDegree(uint32_t node) const {
+    auto local_node = dist_->getLID(node);
+    return graph.getDegree(local_node);
+  }
 
   //! FindEdge searches for the edge between src and dst.
   std::optional<DataGraph::edge_iterator> FindEdge(uint32_t src,
```
l-hoang, line 140: gid

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -117,23 +127,30 @@ class GALOIS_EXPORT AttributedGraph {
   bool MatchNodeDegree(const QueryNode& query, uint32_t node) const;
 
   size_t GetDegree(uint32_t node, const QueryLabel& label) const {
-    return graph.getDegree(node, label);
+    auto local_node = dist_->getLID(node);
+    return graph.getDegree(local_node, label);
   }
 
-  size_t GetDegree(uint32_t node) const { return graph.getDegree(node); }
+  size_t GetDegree(uint32_t node) const {
```
l-hoang, line 134: ditto pretty much fro this entire file

lonestar/libdistbench/include/DistBench/Start.h
```diff
@@ -136,12 +136,14 @@ marshalGPUGraph(DistSubstratePtr<NodeData, EdgeData>& gluonSubstrate,
  * @returns Pointer to the loaded graph
  */
 template <typename NodeData, typename EdgeData, bool IterateOutEdges>
-static DistGraphPtr<NodeData, EdgeData> loadDistGraph() {
+static DistGraphPtr<NodeData, EdgeData>
+loadDistGraph(bool mirror_edges, bool load_all_properties) {
```
l-hoang, line 140: config object for double boolean would prevent errors later

libcusp/include/galois/graphs/DistGraph.h
```diff
@@ -123,6 +123,10 @@ class DistGraph {
     return dist_->Write(rdg_name);
   }
 
+  auto& GetPropertyGraph() const { return dist_->GetPropertyGraph(); }
+
+  auto ReleaseDistribution() { return dist_.release(); }
```
l-hoang, line 128: having the return value be auto here confuses a reader at first glance

is release something that releases the distribution or does it return a "release" object? some documentation would help here as well

l-hoang (commented):
haven't looked at Partitioner.h

.github/workflows/cpp.yaml
```diff
@@ -250,7 +250,7 @@ jobs:
       run: |
         ccache --zero-stats
         ccache --max-files 600
-        cmake --build $HOME/build --parallel 2
+        cmake --build $HOME/build --parallel 1
```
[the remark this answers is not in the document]
roshandathathri: Parallel builds can exceed memory limits and break CI builds, so disabled it.

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -192,7 +163,9 @@ int main(int argc, char** argv) {
   // preload the graph
   galois::StatTimer graph_construct_time("GraphConstructTime");
   graph_construct_time.start();
-  query_server.PreloadGraph(filename);
+  auto dist_graph = loadDistGraph<void, void, true>(true, true);
```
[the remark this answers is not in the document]
roshandathathri: Will replace auto with the type.

lonestar/libdistbench/include/DistBench/Input.h
```diff
@@ -187,7 +187,8 @@ ConstructGraph(PartitioningScheme scheme, bool iterate_out_edges,
  */
 template <typename NodeData, typename EdgeData>
 std::unique_ptr<galois::graphs::DistGraph<NodeData, EdgeData>>
-ConstructGraph(bool iterate_out_edges) {
+ConstructGraph(bool iterate_out_edges, bool mirror_edges,
```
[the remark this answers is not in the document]
roshandathathri: Good point. I'll defer this for later.

libquery/include/galois/graphs/AttributedGraph.h
```diff
@@ -92,7 +97,9 @@ class GALOIS_EXPORT AttributedGraph {
   GetEdgeAttributeArray(const std::string& attribute) const;
 
   bool IsConnected(uint32_t src, uint32_t dst, const QueryLabel& label) const {
```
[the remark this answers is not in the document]
roshandathathri: There's a comment for the class saying in all its APIs, the node identifiers in both input and output are global IDs.

lonestar/libdistbench/include/DistBench/Start.h
roshandathathri: Good point. I'll defer this for later.

libcusp/include/galois/graphs/DistGraph.h
roshandathathri: It returns a unique pointer. There is no release object. Will replace auto with the type.

PR #469, opened 2020-09-02
Return unique_ptr from Table factory methods

ddn0:
Factory methods create new instances of classes. It is more precise to
return a unique_ptr as the caller does not share ownership with the
factory. Callers can weaken their ownership to a shared_ptr themselves.

Along the way, use const references more and fix the schema generation
for empty tables in MatchOp.

Apply some stylistic changes in affected files:
- whitespace between loops, variable declarations and function calls
- reduce nesting level of class definitions


libquery/include/querying/Table.h
```diff
@@ -56,69 +56,36 @@ class GALOIS_EXPORT Table : public InterData {
...
-  Table(Schema schema, std::shared_ptr<arrow::Table> table);
+  static KeyType
+  MakeKey(const std::vector<size_t>& key_indices, const Table::Schema& schema,
+          const std::vector<std::shared_ptr<arrow::Array>>& all_cols,
+          int64_t row_id);
 
-  static std::shared_ptr<Table>
+  //! Create a table from embeddings
+  static std::unique_ptr<Table>
   From(const QueryEmbeddingList& embeddings,
```
roshandathathri, line 69: From -> Make?

libquery/include/querying/Table.h
```diff
@@ -56,69 +56,36 @@ class GALOIS_EXPORT Table : public InterData {
...
-       graphs::AttributedGraph& graph);
+       const graphs::AttributedGraph& graph);
 
-  static KeyType
-  MakeKey(const std::vector<size_t>& key_indices, const Table::Schema& schema,
-          const std::vector<std::shared_ptr<arrow::Array>>& all_cols,
-          int64_t row_id);
+  //! Create an empty table of embeddings
+  static std::unique_ptr<Table>
+  From(const std::vector<TableNodeName>& node_names,
```
roshandathathri, line 76: From->MakeEmpty?

libquery/include/querying/Table.h
```diff
@@ -56,69 +56,36 @@ class GALOIS_EXPORT Table : public InterData {
...
+       const std::vector<TableEdgeName>& edge_names);
 
-  //! Make a new table out of the input table and only keep distinct rows
-  static std::shared_ptr<Table> MakeDistinct(Table*);
+  //! Create a table from another table
+  static std::unique_ptr<Table> From(const Table& input,
```
roshandathathri, line 80: From->Copy?

libquery/include/querying/Table.h
```diff
@@ -56,69 +56,36 @@ class GALOIS_EXPORT Table : public InterData {
...
+                                     const std::vector<bool>& keep_rows);
+
+  //! Create a table from a raw arrow table
+  static std::unique_ptr<Table>
+  From(Schema schema, const std::shared_ptr<arrow::Table>& table);
```
roshandathathri, line 85: From->Make?

libquery/include/querying/Table.h
```diff
@@ -166,6 +133,52 @@ class GALOIS_EXPORT Table : public InterData {
 
   template <typename T>
   std::optional<T> GetElement(size_t index, size_t column_number);
+
+  Table(Schema schema, std::shared_ptr<arrow::Table> table);
+};
+
+//! Class for building table objects manually via Add, from embeddings
```
roshandathathri, line 140: Doesn't Table directly support building from embeddings?

libquery/include/querying/Table.h
```diff
@@ -166,6 +133,52 @@ class GALOIS_EXPORT Table : public InterData {
...
+//! returned by a query, or from a set of arrow arrays
+class Table::Builder {
+  template <typename T>
+  void Add(size_t column_number, T value);
```
roshandathathri, line 144: Can't this be omitted from the header? Why does it need to be in the class?

roshandathathri (approved):
LGTM

Some minor nits about function names

libquery/include/querying/Table.h
ddn0: It can be removed, but it is a little tedious to do so.

The definition of `Add` is simply:

```cpp
template <typename T>
void galois::Table::Builder::Add(size_t column_number, T value) {
  assert(column_number < schema_.NumColumns());
  // check type, respect VERTEX/EDGE
  ArrowAssert(
      ArrowTypes<T>::Specialize(builders_[column_number])->Append(value));
}
```

so two (`schema_`, `builders_`) of its 5 functional dependencies (..., `column_number`, `value`, `T`) are member variables. If I make it a free (but statically linked) function, we'd have to pass these in at the expense of readibility. I figured that just leaving this as a private function is good enough as it prevents anyone directly calling `Add` and avoids my original issue with it, which is that it is only defined for the specific types used in `Table.cpp`.

libquery/include/querying/Table.h
ddn0: Updated the comment.

libquery/include/querying/Table.h
roshandathathri: Makes sense. Thanks.

libquery/include/querying/Table.h
ddn0: Kept it as `Make`. I think `std::unique_ptr<T> Make(const T&)` is good enough for conveying that this is making a copy.

roshandathathri: Ok.

PR #479, opened 2020-09-09
Optimize expression eval and avoid QueryProperty

bowu:
This commit fixes two problems: First, the previous expression evaluation is fundamentally row-based, which does not match the columnar data representation. Second, QueryProperty is not suitable to represent projection expressions.


libquery/src/Expression.cpp
```diff
@@ -0,0 +1,142 @@
...
+  if (left.GetType() == kBoolean) {
+    return BooleanFunc(op, left, right);
+  }
+  // TODO (bowu): use std::visit to traverse all types
+  if ((left.Size() == 1 && right.Size() > 1) ||
+      (left.Size() > 1 && right.Size() == 1)) {
+    Scalar sca          = left.Size() == 1 ? left.ToScalar() : right.ToScalar();
+    const Column& array = left.Size() == 1 ? right : left;
+
+    switch (left.GetType()) {
```
roshandathathri, line 46: Instead of the switch, it might be better to call EvaluateBinaryOp recursively here?

libquery/include/querying/Column.h
```diff
@@ -0,0 +1,89 @@
...
+
+  std::shared_ptr<arrow::Scalar> GetData() const { return value_; }
+};
+
+class Column {
+  TEType type_;
+  std::shared_ptr<arrow::Array> array_;
+
+public:
+  Column() = default;
```
l-hoang, line 57: do you want a default constructor here if there's no way to set the type or array using a public function?

libquery/src/Expression.cpp
```diff
@@ -0,0 +1,142 @@
...
+    Column col = arg->Evaluate(table, graph);
+    vec.push_back(col);
+  }
+  func->SetDataGraph(graph);
+  return func->Go(vec);
+}
+
+galois::Column
+galois::Expression::Evaluate(const Table* table,
+                             const graphs::AttributedGraph* graph) {
```
roshandathathri, line 99: Why pointers instead of const references?

libquery/include/querying/CypherCompiler.h
```diff
@@ -13,14 +13,22 @@
...
-#include "querying/ExpressionTree.h"
+#include "querying/Expression.h"
 
 namespace galois {
 
 using CypherASTNode = const cypher_astnode*;
 using CypherASTType = cypher_astnode_type_t;
 using CypherASTOp   = const cypher_operator_t*;
 
+struct ProjectItem {
```
l-hoang, line 24: i would move this to compiler query structs

libquery/src/CountFunc.cpp
```diff
@@ -0,0 +1,44 @@
...
+  GALOIS_ASSERT(count_result.ok());
+
+  count_datum = std::move(count_result.ValueOrDie());
+#ifdef GALOIS_ENABLE_DIST
+  int64_t count =
+      std::static_pointer_cast<arrow::Int64Scalar>(count_datum.scalar())->value;
+  galois::DGAccumulator<int64_t> total_count;
+  total_count.reset();
+  total_count += count;
+  int64_t count_val = total_count.reduce();
```
roshandathathri, line 32: Setting count instead of count_val will simplify this code so that "else" part is not required.

libquery/include/querying/CypherCompiler.h
```diff
@@ -171,6 +181,28 @@ class GALOIS_EXPORT CypherCompiler {
   //! @param ast Result of some parse from the cypher
   int CompileAST(const cypher_parse_result* ast);
 
+  std::optional<int> GetASTInteger(CypherASTNode number_ast);
```
l-hoang, line 184: if a function doesn't need to access any class variables, they are better off being defined only in the cpp as helpers (which iirc these functions were defined there before this change)

maybe you changed the behavior and they do mess with class vars now; if so ignore this comment

roshandathathri (commented):
Some comments. Haven't seen the 3 largest files yet.

libquery/test/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-add_executable(cypher-compile-test cypher-compile-test.cpp)
-target_link_libraries(cypher-compile-test galois_query)
-add_test(NAME cypher-compile-test COMMAND cypher-compile-test)
-set_tests_properties(cypher-compile-test PROPERTIES LABELS quick)
+#add_executable(cypher-compile-test cypher-compile-test.cpp)
+#target_link_libraries(cypher-compile-test galois_query)
+#add_test(NAME cypher-compile-test COMMAND cypher-compile-test)
+#set_tests_properties(cypher-compile-test PROPERTIES LABELS quick)
```
l-hoang, line 4: unless there's a suitable replacement for these tests, it would be best not to disable them since they exist to test the compiler as a unit separate from everything else

libquery/test/cypher-compile-test.cpp
```diff
@@ -50,20 +50,11 @@ void AssertBasicReturnValue(const galois::CypherCompiler& cc,
                             std::string var_name) {
```
l-hoang, line 50: re: this entire file

it seems like they've been disabled in the cmakelists i saw earlier; if that's the case, do not push any changes you may have since i'm assuming the reason they were disabled was that they were not working

it's best not to push changes that were made in an attempt to get them to work (at least this is what i think this is)

libquery/test/filter-op-test.cpp
```diff
@@ -248,8 +248,10 @@ void ApplyFilterAndCheckNumRows(
     const std::shared_ptr<galois::Table>& test_table,
     std::vector<std::unique_ptr<galois::FilterOp>>&& test_filters,
     size_t expected_num_rows) {
+  int i = 0;
   for (auto&& filter_op : test_filters) {
     // create start node of the op graph then attach the filter node to it
+    std::cerr << "Running filter #" << i++ << "\n";
```
l-hoang, line 254: use GALOIS_LOG_DEBUG or VERBOSE

libquery/test/table-bench.cpp
```diff
@@ -191,20 +191,20 @@ void FilterOp(benchmark::State& state) {
...
 
   galois::Table::Schema schema;
-  schema.AddColumn("test", galois::Table::Type::INT64);
+  schema.AddColumn("test", galois::kInt64);
   std::shared_ptr<galois::Table> table =
       galois::Table::Make(schema, input_table);
 
   std::string query =
-      fmt::format("match (a) where (test + 3) > ({} / 2) return a;", num_rows);
+      fmt::format("match (a) where (test + 3) > {} return a;", num_rows / 2);
```
l-hoang, line 199: does division not work anymore, or was this actually always broken

i can see "always broken" being true because if num_rows isn't even then it definitely wouldn't work

libquery/include/querying/funcs/Func.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "querying/Column.h"
+
+namespace galois {
+
+class Table;
+namespace graphs {
+class AttributedGraph;
+}
+
```
l-hoang, line 14: forward declarations?

libquery/include/querying/funcs/Func.h
```diff
@@ -0,0 +1,49 @@
...
+  void SetDistinct() { distinct_ = true; }
+
+  bool IsDistinct() { return distinct_; }
+
+  virtual bool IsAggregation() { return aggregation_; }
+
+  void SetDataGraph(const graphs::AttributedGraph* graph) { graph_ = graph; }
+
+protected:
+  void SetAggregation() { aggregation_ = true; }
```
l-hoang, line 37: is there a particular reason that the aggregation can't be set when the object is constructed rather than having it (and the data graph) be set separately?

libquery/src/CompilerQueryStructs.cpp
```diff
@@ -34,8 +34,6 @@ void galois::CompilerOrderByMetadata::PrintStruct(unsigned tab_prefix) const {
     } else {
       galois::gPrint("DESC\t");
     }
-    // element on a single line
```
l-hoang, line 37: i would leave a suitable replacement for this removed debug function to keep the original debug functionality there

libquery/src/CountFunc.cpp
```diff
@@ -0,0 +1,44 @@
...
+    GALOIS_ASSERT(unique_result.ok());
+    input_array = std::move(unique_result.ValueOrDie()).make_array();
+  } else {
+    input_array = col_vec[0].GetData();
+  }
+  auto count_result = arrow::compute::CallFunction("count", {input_array});
+  GALOIS_ASSERT(count_result.ok());
+
+  count_datum = std::move(count_result.ValueOrDie());
+#ifdef GALOIS_ENABLE_DIST
```
l-hoang, line 26: i believe DIST is always on now so this ifdef can be removed

libquery/src/FilterOp.cpp
```diff
@@ -8,87 +10,29 @@ std::vector<galois::Operator*> galois::FilterOp::Compute() {
...
-    // may not have properties to substitute (though these should
-    // be optimized away in the optimizer step)
-    if (expression_trees_[0]->Evaluate(map)) {
-      kept_rows[i] = true;
+  Expression* exp = expressions_[0].get();
+  Column col      = exp->Evaluate(input.get(), GetDataGraph());
+  GALOIS_ASSERT(col.GetType() == kBoolean, "Returned array must be boolean");
+
+  if (input->NumRows() > 1 && col.Size() == 1) {
+    // This hack is necessary because of the way we compile "not (p1)-->(p2)"
```
l-hoang, line 20: ?

libquery/src/OpGraphBuilder.cpp
```diff
@@ -101,7 +100,7 @@ DivideCypherAST(cypher_parse_result* cypher_ast_result) {
                      "Expected body of statement node to be a query");
 
   bool currently_parsing_match = false;
-  size_t match_start           = static_cast<size_t>(-1);
+  auto match_start             = static_cast<size_t>(-1);
```
l-hoang, line 103: why auto if you're doing a static case to a size_t

libquery/src/OpGraphBuilder.cpp
```diff
@@ -134,7 +133,9 @@ DivideCypherAST(cypher_parse_result* cypher_ast_result) {
         match_start             = i;
       }
       continue;
-    } else if (child_type == CYPHER_AST_WITH) {
+    }
+
+    if (child_type == CYPHER_AST_WITH) {
```
l-hoang, line 138: why is this not an elseif

a match can't also be a with, no?

l-hoang (commented):
have not looked at cypher compiler or expression changes

also did not look too deeply at semantic changes; most comments are design related

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+#include <unordered_map>
+
+namespace galois {
+
+class Table;
+namespace graphs {
+class AttributedGraph;
+}
+
+struct Identifier {
```
l-hoang, line 27: this name is quite generic and might make it difficult for new devs to understand what purpose it serves

also, it's pretty much just a string at this point; i'm assuming it's to better distinguish something or to make something typesafe

would a type def work better?

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  // construct a single-element column
+  Column ConstructSEArray(ExpressionAbstractType val) {
+    TEType type;
+    std::shared_ptr<arrow::Array> arrow_array;
+    std::visit(
+        [&](auto&& arg) {
+          using T = std::decay_t<decltype(arg)>;
+          if constexpr (std::is_same_v<T, bool>) {
```
l-hoang, line 135: i'm not sure if this is the idiomatic way to use a visitor function; looks kind of clunky

but what do it know i've never used one before :P

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  Expression(std::unique_ptr<Func>&& func,
+             std::vector<std::unique_ptr<Expression>>&& arg_vec) {
+    arith_tree_type_.emplace<kFunction>(std::move(func));
+    function_arguments_ = std::move(arg_vec);
+  }
+
+  // construct a single-element column
+  Column ConstructSEArray(ExpressionAbstractType val) {
```
l-hoang, line 129: i would go ahead and expand SE->SingleElement

way easier for a dev to understand; i initially saw SE and was like "looks complicated"

then i read it and was like "oh this is a singleton column"

libquery/include/querying/Column.h
```diff
@@ -0,0 +1,89 @@
...
+  kInt32,
+  kUInt64,
+  kInt64,
+  kBoolean,
+  kFloat,
+  kDouble,
+  kString,
+  kVertex,
+  kEdge,
+  kInvalid
```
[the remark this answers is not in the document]
ddn0: FWIW, an `enum class` helps alleviate this issue since the names would be referred to as `Foo::kVertex` and never as `kVertex`

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+  //! left part of tree
+  std::unique_ptr<Expression> left_{nullptr};
+  //! right part of tree
+  std::unique_ptr<Expression> right_{nullptr};
+  //! Arguments to functions
+  std::optional<std::vector<std::unique_ptr<Expression>>> function_arguments_;
+
+  template <typename T>
+  Column NumericFunc(ExpressionOperator op, const Column& left,
+                     const T& right) {
```
l-hoang, line 328: why is right a template arg?

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+    auto array                = std::move(result_datum).make_array();
+    return Column(type, std::move(array));
+  }
+
+  template <typename T>
+  Column FloatFunc(ExpressionOperator op, const Column& left, const T& right) {
+    return NumericFunc<T>(op, left, right);
+  }
+
+  Column BooleanFunc(ExpressionOperator op, Column left, Column right) {
```
l-hoang, line 453: since this isn't a template you can move it to a cpp file

though i guess since the rest are here for consistency it probably makes sense to leave this here

l-hoang (commented):
expression.h refactor looks very nice

libquery/include/querying/Column.h
```diff
@@ -0,0 +1,89 @@
...
+#include "galois/gIO.h"
+
+namespace galois {
+
+// Type of a resolved ExpressionAbstractType during runtime
+using ExpressionValue = std::variant<bool, std::string, uint32_t, int32_t,
+                                     uint64_t, int64_t, float, double>;
+
+// table element type
+enum TEType {
```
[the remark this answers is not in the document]
ddn0: I'm not sure in the motivation for establishing a parallel type system when libarrow has its own, including methods for taking a schema of column with nested fields and flattening it into several columns.

How do we plan to support field accesses in nested data types?

libquery/src/Expression.cpp
l-hoang: second'd

this entire function looks menancing when from what i can tell it should just recursively call left right to get values then do the op on them

i might be missing something

libquery/src/Expression.cpp
l-hoang: +1

unless object is being altered

l-hoang (commented):
the only file i haven't looked at at this point is the cyphercompiler

i cannot claim to have looked at groupby in detail because it honestly intimidated me

libquery/src/CypherCompiler.cpp
```diff
@@ -527,103 +382,47 @@ int galois::CypherCompiler::CompilePatternPath(CypherASTNode ast, bool negate) {
...
+    exp_op = kDivide;
   } else if (op == CYPHER_OP_MOD) {
-    clause_op.emplace<kArithmeticTree>(kArithModulo);
+    exp_op = kModulo;
   } else {
-    // return empty
-    GALOIS_LOG_WARN(
-        "Unsupported binary operation; undefined behavior may occur");
-    return CompilerTreeVariant();
+    GALOIS_LOG_FATAL("Unsupported binary operation");
```
l-hoang, line 412: guess you're going for "fail fast" now

libquery/src/CypherCompiler.cpp
```diff
@@ -860,27 +666,53 @@ galois::CypherCompiler::CompilePredicate(CypherASTNode ast) {
...
-    return CompilerTreeVariant(std::in_place_index<kExpressionClause>,
-                               CompileComparison(ast));
+    return CompileComparison(ast);
   } else if (type == CYPHER_AST_UNARY_OPERATOR) {
-    // TODO apparently the negative sign to specify negative numbers is a
-    // unary operator; add support for this
     return CompileUnaryOperator(ast);
-  } else if (type == CYPHER_AST_LABELS_OPERATOR) {
+    //}
+    // else if (type == CYPHER_AST_LABELS_OPERATOR) {
```
l-hoang, line 673: at this point i would just remove this if it's not being used

l-hoang (commented):
quickly skimmed compiler.cpp, looks sane to me

only comment is that the "flow" of the compiler (i.e. what can call what) will be quite confusing to a newcomer; e.g. what is allowed to call compileexpression? are there issues calling it in unexpected places? those kinds of questions

would be good to have it made more clear at some point. from what i can tell it's safe to call compileexpression anywhere

libquery/include/querying/Column.h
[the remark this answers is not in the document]
l-hoang: it's sounding like enum classes are the fix to all the evils of a regular enum

libquery/src/OpGraphBuilder.cpp
l-hoang: realized just a few moments ago that this is because size_t is in the static cast so it's obvious if a dev keeps reading

though my argument is that auto and size_t are a measly 2 characters difference and doesn't sacrifice clarity (in my opinion)

libquery/include/querying/Column.h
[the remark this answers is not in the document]
bowu: It's table element type. I'll use TableElementType.

libquery/src/CypherCompiler.cpp
ddn0: Sometimes this is called the samurai principle: succeed or never return. I actually don't advocate it in general, but it has a better name than expose errors thoughtfully.

libquery/include/querying/Column.h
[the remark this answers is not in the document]
roshandathathri: That seems to be indicate that each element in the table can be of a different type but that isn't the case, right? Each column of the table can have a different type but each element within a column cannot be of different type, right?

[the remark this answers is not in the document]
danielmawhirter: Vertex and Edge types that are distinguishable from uint32 and uint64 are helpful. If we need nested types, I would be in favor of just using the arrow type system. Part of the reason I isolated it in the first place was the assumption that we wouldn't need nested types, and it makes arrow not quite as embedded a dependency.

libquery/include/querying/Column.h
danielmawhirter: If the default were Invalid/None that would be ok, but the current default would effectively be uint8 I believe, since it would be mapped onto 0 in the enum

libquery/include/querying/CypherCompiler.h
```diff
@@ -13,14 +13,22 @@
...
 
 namespace galois {
 
 using CypherASTNode = const cypher_astnode*;
 using CypherASTType = cypher_astnode_type_t;
 using CypherASTOp   = const cypher_operator_t*;
 
+struct ProjectItem {
+  std::unique_ptr<Expression> exp_;
+  std::string alias_;
```
ddn0, line 26: Something that I've been doing informally, but I think is a good convention is to distinguish C-style structs from C++-style classes.

C-style structs are defined with `struct`, callers access fields directly and are expected to maintain type invariants.

C++-style classes are defined with `class`, the class uses constructors and accessors to maintain its own type invariants.

More to the point, it seems reasonable to name the fields of a C-style struct without underscores.

And you can initialize C-style structs directly without a constructor,

```cpp
ProjectItem {
  .exp = ...,
  .alias = ...
};
```

libquery/include/querying/CypherCompiler.h
```diff
@@ -13,14 +13,22 @@
...
+
+  ProjectItem(std::unique_ptr<Expression>&& exp, std::string&& alias)
```
ddn0, line 28: ditto (take by value, then move)

libquery/include/querying/CypherCompiler.h
```diff
@@ -36,20 +44,20 @@ class GALOIS_EXPORT CypherCompiler {
   }
 
   //! Returns set of return structures
-  const std::vector<QueryProperty>& GetReturnValues() const {
-    return this->return_values;
+  std::vector<ProjectItem>&& MoveReturnValues() {
```
ddn0, line 47: Why is this needed?

The value of a moved-from object is valid but implementation-defined.

Moving out individual fields leads to complicated class state (e.g., object with initialized fields, object with some fields in indeterminate state). It is better to move an entire object or none of it (e.g., refactor CypherCompiler to be less monolithic or just accept making a copy).

libquery/include/querying/CypherCompiler.h
```diff
@@ -36,20 +44,20 @@ class GALOIS_EXPORT CypherCompiler {
...
+    return std::move(return_values);
   }
 
   //! Returns structure specifying which return modifiers were specified in
   //! the cypher query
-  const CompilerReturnMetadata& GetReturnMetadata() const {
-    return this->return_metadata;
+  CompilerReturnMetadata&& MoveReturnMetadata() {
```
ddn0, line 53: ditto

libquery/include/querying/CypherCompiler.h
```diff
@@ -134,24 +145,23 @@ class GALOIS_EXPORT CypherCompiler {
   //! Compiles a binary operator and returns a tree for the operator
   //! Binary operation can end up with an expression tree (logical operators)
   //! or an arithmetic tree (arithmetic binary operators)
-  CompilerTreeVariant CompileBinaryOperator(CypherASTNode ast);
+  std::unique_ptr<Expression> CompileBinaryOperator(CypherASTNode ast);
```
ddn0, line 148: maybe `MakeFoo` for these types of factory functions.

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+#include <arrow/compute/api.h>
+#include <memory>
+#include <set>
+#include <variant>
+#include <utility>
+#include <unordered_map>
+
+namespace galois {
+
+class Table;
```
ddn0, line 22: Sometimes forward declarations are necessary but often they are a sign that declarations and classes should be reorganized.

In this case,
- why would an expression need to know what a table is rather than a table evaluating an expression?
- why would an expression need to know what a graph is?

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+namespace graphs {
+class AttributedGraph;
+}
+
+struct Identifier {
+  std::string name_;
+  Identifier(std::string name) : name_(std::move(name)){};
```
ddn0, line 29: ditto my comment about C-structs versus C++ classes.

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+    arith_tree_type_.emplace<kSingleton>(
+        ExpressionAbstractType(std::in_place_index<kEABool>, literal));
+  }
+
+  Expression(const Identifier& id) {
+    arith_tree_type_.emplace<kSingleton>(
+        ExpressionAbstractType(std::in_place_index<kEAIdentifier>, id));
+  }
+
+  Expression(std::unique_ptr<Func>&& func,
```
ddn0, line 122: unnecessary rvalue reference, ditto and throughout.

libquery/include/querying/Expression.h
ddn0: the way to use visit is actually to use function overloading, e.g.,
```cpp

DoStuff(int);
DoStuff(std::string);

std::visit([](auto arg) {
  DoStuff(arg);
}
```

The structure used currently would be most appropriate if what you do upon visiting conceptually different for each type. This, in general, is an unmaintainable pattern because it is difficult to reasonable about what has happened after visiting a variant, and I think if it possible to adopt a more uniform structure it will save much confusion down the line.

Previous context: https://github.com/KatanaGraph/katana/pull/444#discussion_r480832974

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+          using T = std::decay_t<decltype(arg)>;
+          if constexpr (std::is_same_v<T, Identifier>) {
+            names.insert(arg.name_);
+          }
+        },
+        leaf_value);
+    return names;
+  }
+
+  std::string ToString() {
```
ddn0, line 224: if a function is not templated and doesn't need to be inlined, I'd prefer that function definitions be defined out-of-line in a cpp file. This improves compilation times, reduces implementation leakage to other files, allows more freedom in code organization (e.g., the implementation of a class can span multiple files use helper functions with static linkage) and reduces the chance of circular includes.

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+            arrow_array = array;
+          } else {
+            GALOIS_LOG_FATAL("Unsupported type");
+          }
+        },
+        val);
+    return Column(type, std::move(arrow_array));
+  }
+
+  bool HasAggFunc() const {
```
ddn0, line 179: Usually these types of AST aggregators (type propagation, to string, etc.) use an AST visitor pattern where you just apply a user defined function to a generic visitor with a specific traversal order (pre, in, post). 

This might simplify a lot of the code here.

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+      break;
+    case kLessEqual:
+      result = arrow::compute::CallFunction("less_equal",
+                                            {left.GetData(), right.GetData()});
+      type   = kBoolean;
+      break;
+    default:
+      GALOIS_LOG_FATAL("Operator not supported");
+    }
+    GALOIS_ASSERT(result.ok());
```
ddn0, line 388: `GALOIS_LOG_ASSERT` probably

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+    arrow::Datum result_datum = result.ValueOrDie();
+    auto array                = std::move(result_datum).make_array();
+    return Column(type, std::move(array));
+  }
+
+  template <typename T>
```
ddn0, line 394: Why a template here?

as far as I can see, there is only one Scalar type and that is the only argument as `right`.

libquery/src/CypherCompiler.cpp
```diff
@@ -768,37 +553,37 @@ galois::CypherCompiler::CompileUnaryOperator(CypherASTNode ast) {
...
   }
-
-  GALOIS_LOG_WARN("An unsupported unary operation has been detected. Ignoring "
-                  "it. Undefined behavior may occur");
-  // empty node; shouldn't get here at this moment in time
-  return CompilerTreeVariant();
+  GALOIS_LOG_FATAL("Only NOT is supported");
 }
 
+//   GALOIS_LOG_WARN("An unsupported unary operation has been detected.
```
ddn0, line 579: delete?

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,497 @@
...
+  std::string name_;
+  Identifier(std::string name) : name_(std::move(name)){};
+};
+
+//! Type of an unevaluated expression stored in an expresson tree. Upon
+//! evaluation, it resolves into a more concrete type.
+using ExpressionAbstractType =
+    std::variant<Identifier, std::string, int64_t, float, bool>;
+//! Enum to make indexing into ExpressionAbstractType more readable
+enum ExpressionAbstractIndex {
```
ddn0, line 37: I'd prefer that we have better visiting patterns than having to rely on these index enums (throughout). They are pretty hacky.

ddn0 (commented):
I think evaluating the expressions column-wise is a great idea.

This has also raised some issues around the interaction between ASTs, Tables, Columns, etc that need to worked through.

I don't mind deferring some of this work to a latter PR, but at the very least we should keep the test coverage the same (or better).

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,324 @@
...
+               std::unique_ptr<Func>>
+      arith_tree_type_;
+  //! left part of tree
+  std::unique_ptr<Expression> left_{nullptr};
+  //! right part of tree
+  std::unique_ptr<Expression> right_{nullptr};
+  //! Arguments to functions
+  std::optional<std::vector<std::unique_ptr<Expression>>> function_arguments_;
+
+  class ExpressionVisitor {
```
roshandathathri, line 164: Can't this moved to the cpp file?

libquery/include/querying/Expression.h
```diff
@@ -0,0 +1,324 @@
...
+    Column operator()(ExpressionAbstractType& val);
+
+    Column operator()(ExpressionOperator& op);
+
+    Column operator()(std::unique_ptr<Func>& func);
+  };
+
+  template <typename T>
+  Column NumericFunc(ExpressionOperator op, const Column& left,
+                     const T& right) const {
```
roshandathathri, line 183: Private template functions don't need to be defined in the header, so move it to the cpp file (throughout).

roshandathathri (approved):

libquery/CMakeLists.txt
```diff
@@ -34,7 +35,9 @@ target_include_directories(galois_query PUBLIC
 
 target_link_libraries(galois_query PRIVATE ${Boost_LIBS} cypherparser)
 target_link_libraries(galois_query PUBLIC Galois::shmem)
-target_link_libraries(galois_query PUBLIC Galois::cusp)
+if(GALOIS_ENABLE_DIST)
```
roshandathathri, line 38: Drop if. galois_query is only built if GALOIS_ENABLE_DIST.

libquery/test/ExecutionPlanTest.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  galois::SharedMemSys G;
+
+  galois::StatTimer total_time("TimerTotal");
+  total_time.start();
+
+  galois::QueryServer server;
+  server.RunCypherQuery(graph_file.getValue(), query);
```
roshandathathri, line 57: Why is this needed? graph-query in lonestar does the same thing, right?

l-hoang: this goes back to the point Donald made of "something end-to-end" vs "unit-test"

he can explain it much better than i can

ddn0: My point was that we should have the infrastructure for both; though at this point in time, we only have "end-to-end" tests. If `ExecutionPlanTest.cpp` is just going to have the same functionally as `graph-query` there is no point in having both, which I think is @roshandathathri 's point.

roshandathathri: Yes, I was concurring with Donald's earlier point. libquery/test should have unit tests but this seems liked an end-to-end test that replicates graph-query behavior.

PR #492, opened 2020-09-16
tsuba: use name/version service to identify RDGs

tylershunt:
This is a precursor to another PR coming soon which will connect to an
external name/version service.

Visible changes from to the graph engine: users of tuba should no longer
pass the name of a "meta file"; rather they should consider the
directory that the graph is stored in as the name of the graph. Doing
so, graphs can be updated in place and tsuba will manage the meta files
under the hood to ensure that latest version is read.

Partially to support some existing tests and partially to lay the
groundwork for loading old versions of the graph, I added new interfaces
to `Load` take a uri instead of a handle, these functions accept the
either graph names or meta files. My intention is to move towards a
world where holding a `handle` corresponds to holding a lock on the RDG.
So these interfaces could be used to read a graph that you intend to
write in a new location or read an old version of a graph that you
intend to write back as the latest version, allowing the implementation
of time travel.

There are a lot of churn-y clean up changes especially in RDG which I
thought were important, and definitely made reasoning about the
functionality changes easier.


libtsuba/include/tsuba/RDG.h
```diff
@@ -121,6 +154,66 @@ struct GALOIS_EXPORT RDG {
...
+  LoadPartial(RDGHandle handle, const SliceArg& slice,
+              const std::vector<std::string>* node_props = nullptr,
+              const std::vector<std::string>* edge_props = nullptr);
+  static galois::Result<RDG>
+  LoadPartial(const std::string& rdg_meta_path, const SliceArg& slice,
+              const std::vector<std::string>* node_props = nullptr,
+              const std::vector<std::string>* edge_props = nullptr);
+
+private:
+  galois::Result<void> DoLoad(const std::string& metadata_dir);
```
ddn0, line 192: could you add some newlines around here? my brain can't parse return types versus member fields versus function definitions fast enough.

libtsuba/src/NameServerClient.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+} // namespace tsuba
+
+galois::Result<std::unique_ptr<tsuba::NameServerClient>>
+tsuba::ConnectToNameServer() {
+  std::string host;
+  int port = 0;
+
+  galois::GetEnv("GALOIS_NS_HOST", &host);
```
ddn0, line 91: is the plan to always configure the nameserver via environment variables versus some way of passing this info via tsuba::Init and co?

libtsuba/src/RDG.cpp
```diff
@@ -52,144 +53,57 @@ static const char* global_to_local_vals_prop_name = "global_to_local_values";
...
+  return !std::regex_match(uri, kMetaFileRegex);
 }
 
-} // namespace tsuba
+const std::regex kMetaVersion("meta(?:_([0-9]+)|)");
 
-namespace {
+galois::Result<uint64_t> Parse(const std::string& str) {
+  uint64_t val = strtoul(str.c_str(), NULL, 10);
+  if (val == ULONG_MAX && errno == ERANGE) {
```
ddn0, line 92: maybe just `if (errno == ERANGE)` ? The standard says that `ULONG_MAX` will be returned but it reads a little strange if you are not familiar with the standard because presumably this should be an error whenever errno is ERANGE regardless of the return value.

https://en.cppreference.com/w/c/string/byte/strtoul

libtsuba/src/RDG.cpp
```diff
@@ -1001,99 +691,47 @@ BindOutIndex(const std::string& topology_path) {
...
   galois::Result<void> ret = galois::ResultSuccess();
 
   TSUBA_PTP(tsuba::internal::FaultSensitivity::High);
   comm->Barrier();
+
+  // NS handles MPI coordination
+  if (auto res = tsuba::NS()->Update(handle.impl_->rdg_meta.dir_,
+                                     handle.impl_->rdg_meta.version_, new_meta);
+      !res) {
+    GALOIS_LOG_ERROR("nameserver UpdateRDGMeta {}: {}", res.error(),
```
ddn0, line 709: nit: `GALOIS_LOG_ERROR("unable to update file at {}: {}", handle.impl_->rdg_meta.dir_, res.error())` or similar. I find that putting the context we know first then appending the context from the returned call is a little easier to parse even if you are not familiar with the subsystem.

libtsuba/src/RDG.cpp
```diff
@@ -1105,414 +743,881 @@ galois::Result<void> CommitRDG(tsuba::RDGHandle handle, uint32_t policy_id,
...
+    return tsuba::ErrorCode::InvalidArgument;
+  }
+  return tsuba::RDGMeta::PartitionFileName(meta.dir_, tsuba::Comm()->ID,
+                                           meta.version_);
+}
+
+galois::Result<std::string> GetPartPath(const std::string& rdg_meta_uri,
+                                        bool intended_partial_read) {
+  auto rdg_res = tsuba::RDGMeta::Make(rdg_meta_uri);
+  GALOIS_LOG_DEBUG("{} after GetPartPAth", rdg_meta_uri);
```
ddn0, line 774: needed?

libtsuba/src/RDG.cpp
```diff
@@ -1105,414 +743,881 @@ galois::Result<void> CommitRDG(tsuba::RDGHandle handle, uint32_t policy_id,
...
+    if (res) {
+      res.value().dir_ = rdg;
+    }
+    return res;
+  }
+  return MakeFromStorage(rdg);
+}
+
+std::string RDGMeta::ToJsonString() const {
+  // POSIX specifies that test files end in a newline
```
ddn0, line 907: :D

ddn0 (commented):
partial pass.... I can finish up later if you want more comments.

libtsuba/include/tsuba/RDG.h
```diff
@@ -94,25 +119,33 @@ struct GALOIS_EXPORT PartitionMetadata {
...
-  std::shared_ptr<arrow::Table> node_table;
-  std::shared_ptr<arrow::Table> edge_table;
+  std::shared_ptr<arrow::Table> node_table_;
+  std::shared_ptr<arrow::Table> edge_table_;
 
-  std::vector<tsuba::PropertyMetadata> node_properties;
-  std::vector<tsuba::PropertyMetadata> edge_properties;
-  std::vector<tsuba::PropertyMetadata> part_properties;
-  std::vector<std::pair<std::string, std::string>> other_metadata;
+  std::vector<tsuba::PropertyMetadata> node_properties_;
```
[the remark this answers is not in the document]
tylershunt: Yes that's the plan.

libtsuba/src/NameServerClient.cpp
tylershunt: I plan to eventually move to passing the name server client in a la the `CommBackend`.

libtsuba/include/tsuba/RDG.h
tylershunt: You got it!

libtsuba/src/RDG.cpp
tylershunt: 👍

libtsuba/src/RDG.cpp
ddn0: now that I read it, it seems like s/test/text/

tylershunt: Oh man I didn't even see that.

tylershunt:
Hmm my builds keep getting cancelled even though the only actual failures appear to be in non-required checks.

ddn0 (approved):

libcusp/include/galois/graphs/DistGraph.h
```diff
@@ -340,21 +340,22 @@ PartitionGraph(const DistOptions& opt) {
...
-  } else if (stat.num_hosts == net.Num && stat.transpose == opt.transpose &&
-             p_id == Policy::ID) {
+  if (stat.num_hosts == net.Num && stat.transpose == opt.transpose &&
+      p_id == Policy::ID) {
     if (auto res = dist->LoadFromStorage(opt.filename); !res) {
       GALOIS_LOG_FATAL("failed to populate distribution: {}", res.error());
     }
+  } else if (p_id == PartitionPolicyID::kNotPartitioned ||
+             stat.num_hosts == 1) {
+    assert(stat.num_hosts == 1);
```
roshandathathri, line 356: This assertion seems weird.

libgalois/include/galois/OutIndexView.h
```diff
@@ -27,6 +27,7 @@ class GALOIS_EXPORT OutIndexView {
   OutIndexView& operator=(OutIndexView&& other) = default;
 
   static galois::Result<OutIndexView> Make(tsuba::RDGHandle handle);
```
roshandathathri, line 29: Is this one still used?

libgalois/include/galois/PartialGraphView.h
```diff
@@ -45,65 +45,65 @@ class PartialGraphView {
...
+public:
+  typedef StandardRange<boost::counting_iterator<uint64_t>> edges_iterator;
+  typedef StandardRange<boost::counting_iterator<uint64_t>> nodes_iterator;
 
-    auto rdg_res = tsuba::LoadPartial(handle, node_range, edge_range,
-                                      edges_start, edges_stop - edges_start);
+  /// Make a partial graph view from a partially loaded RDG, as indicated by a
+  /// RDGHandle and OutIndexView, which loads only the specified properties
+  static galois::Result<PartialGraphView>
+  Make(tsuba::RDGHandle handle, OutIndexView&& oiv, uint64_t first_node,
```
roshandathathri, line 74: Is this still used?

Also, what happened to the variant that loads all properties? Does nullptr assume load all properties? If so, there is no option to load none of the properties.

libcusp/include/galois/graphs/DistGraph.h
tylershunt: If the graph is `kNotPartitioned` `stat.num_hosts` should be `1`. A sanity check.

libgalois/include/galois/PartialGraphView.h
tylershunt: `nullptr` assumes that all properties are loaded, non-null pointers to empty vectors would load none of the properties.

libgalois/include/galois/OutIndexView.h
tylershunt: It is not but I think it probably will be; I the plan is for holding a handle to mean holding a lock on the graph, so if you want to load a graph and hold a lock while you write it you would use this variant.

libgalois/include/galois/PartialGraphView.h
roshandathathri: Makes sense.

libcusp/include/galois/graphs/DistGraph.h
roshandathathri: Ok.

PR #493, opened 2020-09-17
Print useful environment variables in for dist execution

ddn0:


libdist/src/DistStats.cpp
```diff
@@ -28,29 +28,24 @@
...
   assert(d && "dynamic_cast<DistStatManager*> failed");
 
   return d;
 }
 
-inline static DistStatManager*
-dsm(void) {
-  return internal::distSysStatManager();
-}
+galois::runtime::DistStatManager::DistStatManager() = default;
```
roshandathathri, line 46: Just curious. This is better suited for the header right?

roshandathathri (approved):
LGTM

libdist/src/DistStats.cpp
ddn0: Yeah; trivial constructors are well-suited to being defined in headers, especially for classes that will be constructed a lot.

Some common cases where you would want to define constructors out-of-line:
- If the class is using the private implementation idiom https://en.cppreference.com/w/cpp/language/pimpl and you are using a unique_ptr to hold the private implementation. In this case, the destructor needs to be defined out-of-line.
- A long constructor body; this case should be rare because constructors shouldn't be that complicated or you should create a separate initialization function 

Statistics/DistStats is surprisingly complicated for something that computes a min, max, average on key-value maps. I tried to limit this set of changes to things that would limit its blast radius rather than improving much internally.

PR #494, opened 2020-09-18
Update git blame ignore

ddn0:


arthurp (approved):
LGTM. The conda recipe reformat is just fine. Other changes look just fine, but I only spot checked.

PR #496, opened 2020-09-18
build: record git sha so that we can write it into binaries

tylershunt:
Having cmake config depend on `.git/index` ensures that the current
sha will be recorded whenever git state changes.

Also a fallback to record "unknown" if for some weird reason git is not
installed.


cmake/Modules/GitHeadSHA.cmake
```diff
@@ -0,0 +1,17 @@
+find_package(Git QUIET)
+
+if (Git_FOUND)
+execute_process(
```
ddn0, line 4: can you append something to indicate if the build is from a clean or dirty working directory?

tylershunt: It's more complicated to get the working directory state without forcing a re-configure on every build. I'm working that up in my free time.

ddn0: I don't think it's worth complicating the build to get a "dirty" flag in.

We probably can get the same value by doing the complement and just having a way to specially mark release builds (e.g., CI conda builds) FWIW.

I'm okay with merging as is.

tylershunt:
I added the build type (e.g. `f6c9d19a9-release`) because that seemed like a good idea too.

ddn0 (approved):

PR #500, opened 2020-09-22
Pass sync structures by value

ddn0:
After #498 

```
Sync structures are passed as template parameters but this prevents sync
methods from having any non-global state (e.g., stack variables).
```


roshandathathri (approved):
This looks great!

PR #508, opened 2020-09-23
Use out-of-line definitions for Distribution

ddn0:
In my testing, this doesn't have a measurable impact on our build times
but it is good practice nevertheless.


roshandathathri (approved):
LGTM

gurbinder533 (approved):

tylershunt (approved):

witchel (approved):
I'm super excited for this change.  While the compile time of Distribution.cpp might be the same as Distribution.h, this change should decrease the total compile time for all the programs that include Distribution.h, right?

ddn0:
> I'm super excited for this change. While the compile time of Distribution.cpp might be the same as Distribution.h, this change should decrease the total compile time for all the programs that include Distribution.h, right?

In principle, yes; though the benefit probably won't be noticeable. But a journey begins with its first step.

PR #509, opened 2020-09-24
Rename PartitioningPolicies.h to PartitionPolicies.h

ddn0:
The word "partitioning" doesn't appear in Partition{ing}Policies.h at
all and the file provides definitions of PartitionPolicyID and
PartitionPolicyType. It seems reasonable to drop the {ing} for some
consistency.


roshandathathri (approved):
LGTM

gurbinder533 (approved):

witchel:
It was annoying when switching buffers in emacs.  Partition is a better prefix.

PR #514, opened 2020-09-28
name-service: connect to etcd if configured

tylershunt:
The name service now operates in two modes: the first is the old in-memory mode which technically provides the transactional guarantee that we claim (with a single point of failure and the whole "not-durable" thing) for testing.

In the second mode, it uses etcd to store changes. Etcd can run on multiple nodes giving us a more palatable story about failures and also logs to the local fs so there is some durability guarantee even if all hosts crash (in practice it would log to elastic block store, so our durability guarantees reduce to theirs).

In the second mode, the name server instances that the katana graph engine talks to have no application state so we could deploy an instance of the name server on every host running graph computation (if the extra network hop turns out to be a problem).

I had to fiddle with the MPI id checking in `HttpNameServerClient` to stop tests from blocking forever when host 0 fails a check, this happened for instance when running the tests twice without dumping server state (create is necessarily only carried out by host 0 and fails if what you're creating already exists). Rather than address that problem fully here (I now MIP abort instead) I plan to fix both MPI error propagation for these cases and fix the tests to properly clean up this state so that re-running without restarting the server is supported in my next PR.

After #506  


tylershunt:
Some more detail: I'm using the etcdv3 `transaction` operation to implement an atomic test and set. Each RDG is actually stored as 2  values: a `version` value that holds the current version and an `RDGMeta` value that holds the meta. I need both because I'd like to be able to return the Meta struct to the application (saving one round trip to s3) and because the comparison operations etcd gives me in `transaction` are lexicographical comparisons of the whole value (afaict). Storing the version separately allows me to do an easy comparison between the application's expected transaction value and the stored current version of the RDG.

name-service/internal/rdg/rdg.go
```diff
@@ -90,3 +98,81 @@ func (ms *memoryStore) TestAndSet(name string, meta Meta, expectedVersion int) e
...
+	}
+	if err = json.Unmarshal(gresp.Kvs[0].Value, &meta); err == nil {
+		return meta, errors.Wrap(err, "Meta unmarshal failed")
+	}
+	return meta, nil
+}
+
+func versionKey(name string) string {
+	return name + ";ver;"
+}
```
witchel, line 133: What does this key mean?

name-service/internal/rdg/rdg.go
```diff
@@ -35,9 +40,9 @@ type Meta struct {
 }
 
 type Store interface {
-	Get(name string) (Meta, error)
-	Create(name string, meta Meta, overwrite bool) error
-	TestAndSet(name string, meta Meta, expectedVersion int) error
+	Get(ctx context.Context, name string) (Meta, error)
+	Create(ctx context.Context, name string, meta Meta) error
```
witchel, line 44: What is the context object?

witchel (approved):
My go fu is weak, so just two questions.

name-service/internal/rdg/rdg.go
tylershunt: 2 keys per RDG; I'm leaving the Key for the Meta struct as whatever the user passed in. This is the key for it's corresponding version value that I use for easy lexicographical comparisons. I appended `;ver;` because it seemed unlikely to appear in a path and a name that we could make illegal but it's easy to change if anyone has ideas.

name-service/internal/rdg/rdg.go
tylershunt: Context objects are for timeouts/canceling ops in Go. Basically the context owner (in this case usually `Gin` the package managing http connections) can tell us to stop what were doing by signaling us through the context. Not so important for in-memory ops but the etcd ops could all block for a while and take a context argument so passing the context from the request seemed the right thing to do here.

name-service/internal/rdg/rdg.go
ddn0: small nit about metadata, reserved names and escaping, which may or may not apply here: as you can tell from the original naming of the RDG metadata keys, I'm a big fan of establishing some namespaced schema for keys and reserved keys (kg.v1.etc) early or just baseXY-encoding user data if you want to leave it uninterpreted.

tylershunt: That makes sense, something like:
`kg.v1.graph.name/<base64encodedname>`
`kg.v1.graph.version/<base64encodedname>`
definitely seems better than what I've got.

ddn0: yeah. base64 (url-encoding variant) so we don't have `/` in the output. We are now ready for emoji names!

tylershunt:
Updated with base64 encoding.

PR #527, opened 2020-10-02
Parameter support

bowu:
This commit supports parameters in both matching and expression evaluation.

Examples:

`mpiexec -np 1 ./bin/graph-query ldbc_003 --query "match (p:Person {id:\$id})-->(p1:Person {id:\$id1}) return p.firstName, p1.id" -parameter=id=21,id1=17592186044420`

`mpiexec -np 1 ./bin/graph-query ldbc_003 --query "match (p:Person) where p.id<\$id return p.firstName" -parameter=id=22`


libquery/src/Expression.cpp
```diff
@@ -11,6 +11,72 @@
 #include "querying/Column.h"
 #include "querying/Table.h"
 
+namespace {
+
+bool
+IsInteger(const std::string& str) {
+  try {
```
ddn0, line 18: https://en.cppreference.com/w/cpp/utility/from_chars might be more enjoyable because it is non-locale dependent and non-throwing.

Also, why not just return the parsed value as a std::optional? re-parsing allows for subtle errors to be introduced, e.g., IsInteger may trim whitespace but the subsequent re-parsing function doesn't.

libquery/include/querying/Expression.h
```diff
@@ -77,6 +60,59 @@ enum ExpressionOperator {
   kLessEqual
 };
 
+extern std::optional<int64_t> ToInteger(const std::string& str);
+
+extern std::optional<double> ToFloat(const std::string& str);
```
roshandathathri, line 65: Why is extern required for these function definitions?

libquery/include/querying/Expression.h
```diff
@@ -150,10 +193,23 @@ class Expression {
...
 
   bool IsValid() const {
-    if (arith_tree_type_.index() == kOperator && !left_) {
+    std::optional<bool> result;
+    std::visit(
+        [&](auto&& arg) {
+          using T = std::decay_t<decltype(arg)>;
+          if constexpr (std::is_same_v<T, ExpressionOperator>) {
+            if (!left_) {
+              result = false;
```
roshandathathri, line 207: Can't return directly from here?

libquery/src/CypherCompiler.cpp
```diff
@@ -76,19 +70,21 @@ GetPropertiesFromMap(CypherASTNode map_ast) {
...
-      property_value.emplace<kEVInt64>(
+      property_value.emplace<int64_t>(
           std::stol(cypher_ast_integer_get_valuestr(ast_value)));
     } else if (value_type == CYPHER_AST_FLOAT) {
-      property_value.emplace<kEVFloat>(
-          std::stof(cypher_ast_float_get_valuestr(ast_value)));
-      // TODO handle parameter
-      // } else if (value_type == CYPHER_AST_PARAMETER) {
-      //   property_value.emplace<3>(cypher_ast_parameter_get_name(ast_value));
+      property_value.emplace<double>(
```
roshandathathri, line 82: Default float type is double?

libquery/src/CypherCompiler.cpp
```diff
@@ -184,7 +177,8 @@ galois::CypherCompiler::CompileNodePatternPath(
   // parse property map
   CypherASTNode property_map_node =
       cypher_ast_node_pattern_get_properties(element);
-  mn.property_map = GetPropertiesFromMap(property_map_node);
+  mn.property_map = GetProperties(property_map_node).first;
+  mn.parameters = GetProperties(property_map_node).second;
```
roshandathathri, line 181: `[mn.property_map, mn.parameters] = GetProperties(property_map_node)` doesn't work?

libquery/src/CypherCompiler.cpp
```diff
@@ -306,7 +300,8 @@ galois::CypherCompiler::CompileRelPatternPath(CypherASTNode element) {
   CypherASTNode property_map_node =
       cypher_ast_rel_pattern_get_properties(element);
   this->query_edges.back().property_map =
-      GetPropertiesFromMap(property_map_node);
+      GetProperties(property_map_node).first;
+  this->query_edges.back().parameters = GetProperties(property_map_node).second;
```
roshandathathri, line 304: Ditto

roshandathathri (commented):
Please add tests to graph-query for parameters - you can create a new querylist for parameter based tests.

Some minor nits below

libquery/src/CypherCompiler.cpp
ddn0: or `std::tie(mn.property_map, mn.parameters) = GetProperties(property_map_node)`

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -80,17 +80,40 @@ ProcessQueryFile(
...
   // putting into string stream lets you pull a string out of it
   query_ss << query_stream.rdbuf();
 
-  if (!query_stream) {
-    GALOIS_LOG_FATAL("failed to read query from string");
+  std::string param_file_name =
+      file_to_process.substr(0, file_to_process.find('.')) + ".param";
+  std::ifstream param_stream(param_file_name);
+  if (param_stream) {
+    galois::gInfo("Reading query file ", file_to_process);
```
roshandathathri, line 96: param file

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -71,7 +71,7 @@ ProcessQueryFile(
     galois::QueryServer* query_server, std::string graph_name,
     std::string file_to_process,
     // TODO bring back parameter map
-    const std::unordered_map<std::string, std::string>&,
+    std::unordered_map<std::string, std::string>& parameter_map,
```
roshandathathri, line 74: It is either `const T&` or `T*` according to Katana coding convention.

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -80,17 +80,40 @@ ProcessQueryFile(
...
   query_ss << query_stream.rdbuf();
 
-  if (!query_stream) {
-    GALOIS_LOG_FATAL("failed to read query from string");
+  std::string param_file_name =
+      file_to_process.substr(0, file_to_process.find('.')) + ".param";
+  std::ifstream param_stream(param_file_name);
+  if (param_stream) {
+    galois::gInfo("Reading query file ", file_to_process);
+    parameter_map.clear();
```
roshandathathri, line 97: This is strange. Why is this passed in as a parameter to this function? Why can't parameter_map just be a local variable?

lonestar/querying/distributed/graph-query/CMakeLists.txt
```diff
@@ -86,6 +86,21 @@ foreach (np ${partitions})
       ${CMAKE_CURRENT_SOURCE_DIR}/cypher-queries/003-dist-querylist.txt.truth
       ${output_location}/queries_p${np}_t${thr}.count)
 
+  add_test(NAME run-dist-graph-query-param-${np}
```
roshandathathri, line 89: I understand this can be confusing. Please add run-graph-query-param* as well above. That will be run on 1 host only.

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -80,17 +79,40 @@ ProcessQueryFile(
...
 
-  if (!query_stream) {
-    GALOIS_LOG_FATAL("failed to read query from string");
+  std::string param_file_name =
+      file_to_process.substr(0, file_to_process.find('.')) + ".param";
+  std::ifstream param_file(param_file_name);
+  if (param_file) {
+    galois::gInfo("Reading query file ", file_to_process);
+    std::string line;
+    while (std::getline(param_file, line)) {
```
roshandathathri, line 98: Maybe better to have it in a separate function but if you have no other changes, it's fine to merge as-is.

lonestar/querying/distributed/graph-query/graph-query.cpp
```diff
@@ -242,13 +261,16 @@ main(int argc, char** argv) {
...
     // single query file
-    ProcessQueryFile(
-        &query_server, input_file, query_base_path + query_file, parameter_map);
+    ProcessQueryFile(&query_server, input_file, query_base_path + query_file);
   } else if (!query.empty()) {
     // query passed in via CL
     galois::StatTimer timer("Query_Timer");
+    // holds parameter values for cypher query passed in via command line
+    std::unordered_map<std::string, std::string> parameter_map =
+        ParseParameters(host_id);
```
roshandathathri, line 270: Slightly confused by this. So if the query is passed via CL, parameters are not supported (because parameters are in a file)?

I think it's useful to distinguish b/w file and CL for both query and parameters. That is:
- query can be in a file or passed via CL
- parameters can be in a file or passed via CL.

However, it's fine to say that parameters can be only specified in a file. In which case, it should work whether the query is in a file or the CL.

lonestar/querying/distributed/graph-query/CMakeLists.txt
```diff
@@ -36,6 +36,21 @@ foreach (thr ${threads})
...
       ${output_location}/queries_p1_t${thr}.count)
 
+  add_test(NAME run-graph-query-param-${thr}
+    COMMAND graph-query
+      ${ldbc_input}
+      -queryBasePath=${samples}
+      -listOfQueries=${samples}/querylist-param.txt
+      -t=${thr}
+      -output
+      -outputLocation=${output_location})
```
roshandathathri, line 46: Note that if PR #526 is merged before this, then the value for this command line option needs to be changed.

I don't expect these tests to pass on CI because run-graph-query-param-${thr} and run-graph-query-${thr} can run in parallel and both write to the same output location. I saw similar issues for #526, so I fixed the issue by changing the meaning of outputLocation. Check graph-query.cpp and it's CMakeLists.txt in #526.

roshandathathri (approved):
LGTM

Minor nits below

PR #529, opened 2020-10-02
Clean up naming of sync structure functions

ddn0:
Mostly because the interface includes an outlier name, setVal_batch,
which is neither here nor there, I decided to update all the sync
structure methods to follow a single well-known naming convention.

I left the bitset functions as is because that change would require
updating more code outside of Gluon.


roshandathathri (approved):
LGTM

gurbinder533 (approved):
LGTM.

PR #531, opened 2020-10-02
Generalize file system testing for Azure

witchel:
I figured Tyler shouldn't be the only one to play with Azure, so he helped me get an account and I wanted to run some tests.  Our tests are a little S3-centric so I made them less so and streamlined them a bit.  While I was at it, I thought I might as well get rid of FileAsyncWork in favor of std::future.

I test for future validity in the code because the documentation told me to.

The whole thing was feeling a little indulgent, but then I found a small bug in our azure code that would crash any attempts to write a file that was 15 bytes or fewer which is the kind of thing these tests are supposed to alert us about.


libtsuba/include/tsuba/file.h
```diff
@@ -50,8 +50,8 @@ GALOIS_EXPORT galois::Result<void> FileStore(
     const std::string& uri, const uint8_t* data, uint64_t size);
 
 // Take whatever is in @data and start putting it a the file called @uri
-GALOIS_EXPORT galois::Result<std::unique_ptr<tsuba::FileAsyncWork>>
-FileStoreAsync(const std::string& uri, const uint8_t* data, uint64_t size);
+GALOIS_EXPORT galois::Result<std::future<galois::Result<void>>> FileStoreAsync(
+    const std::string& uri, const uint8_t* data, uint64_t size);
```
tylershunt, line 54: A suggestion: I have some preference for making these functions return `std::future<galois::Result<void>>` then reporting any errors in the single `galois::Result<void>`, for purely ergonomic reasons.

I still think what's here is better than what come before though.

libtsuba/src/FileView.cpp
```diff
@@ -457,12 +457,12 @@ FileView::Resolve(int64_t start, int64_t size) {
     if (fetch->first_page <= page_number(start + size) ||
         fetch->last_page >= page_number(start)) {
       // Complete the remaining work if there is some
-      if (fetch->work != nullptr) {
-        while (!fetch->work->Done()) {
-          if (auto res = (*fetch->work)(); !res) {
-            return res.error();
-          }
+      if (fetch->work.valid()) {
```
tylershunt, line 460: I would think this would only return `false` on a double get (or some other bug). Maybe make this an `assert` instead? I think making the API contract "These functions alway return valid futures" is more attractive than having to do this nested checking everywhere.

libtsuba/src/RDG.cpp
```diff
@@ -751,11 +751,22 @@ CommitRDG(
...
+            tsuba::RDGMeta::FileName(
+                handle.impl_->rdg_meta.dir_, new_meta.version_),
+            res.error());
+        return res.error();
       }
+    } else {
+      GALOIS_LOG_ERROR(
+          "future invalid in commitRDG {}",
+          tsuba::RDGMeta::FileName(
+              handle.impl_->rdg_meta.dir_, new_meta.version_));
```
tylershunt, line 768: Nit this error was DEBUG level the other time you wrote it out. This is assuming you don't agree that we should make them `asserts`

libtsuba/src/azure.cpp
```diff
@@ -114,9 +114,11 @@ tsuba::AzureGetSync(
 
   std::unique_ptr<az::blob_client> client = std::move(client_res.value());
 
+  // set parallelism to size to work around azure library's dubious:
+  //   block_size = size / parallelism;
   auto ret = client
                  ->download_blob_to_buffer(
-                     container, blob, start, size, result_buf, INT_MAX)
+                     container, blob, start, size, result_buf, size)
```
tylershunt, line 121: oh neat!

tylershunt (approved):
Sweet, thanks for the Azure love.

tylershunt:
> I test for future validity in the code because the documentation told me to.

I assume you're talking about this bit from https://en.cppreference.com/w/cpp/thread/future/get ?
```
The behavior is undefined if valid() is false before the call to this function.

Any shared state is released. valid() is false after a call to this method.
```
I took that to mean the future should be in a state where `valid()` would return true, not that the check was necessary. I *think* a future returned from `std::async` is guaranteed to be valid but I'm far from an expert here.

libtsuba/include/tsuba/file.h
witchel: I'm totally on board.  There is a powerful symmetry about everything returning `galoisResult`, but it is also delightful to get a break, like in `FileSystem.h` or some `Uri` functions.

libtsuba/src/FileView.cpp
witchel: Yeah, I called attention to it because I don't like it either.  Assert seems like the right idiom.

libtsuba/src/azure.cpp
witchel: I'd argue that this is an Azure interface library bug, but we should work around it.

libtsuba/src/RDG.cpp
witchel: Changed in the new version.

libtsuba/src/azure.cpp
tylershunt: Yeah the library isn't great but I think the cpplite library still attractive from a dependency management perspective.

PR #533, opened 2020-10-05
tsuba: add api call to "forget" graphs

tylershunt:
We need some mechanism to let the name server know that we've deleted
a graph, this case comes up a lot in testing where the pre-test
operation clears output of a past test run, for example.

This is my attempt at such a mechanism.

After #532, the changes for this PR are in the last commit.


libtsuba/include/tsuba/RDG.h
```diff
@@ -293,7 +293,11 @@ GALOIS_EXPORT galois::Result<void> Create(const std::string& name);
 /// Register a previously created RDG attaching it to the namespace; infer the
 /// version by examining files in name
 /// \param name is storage location prefix that the RDG is stored in
-galois::Result<void> Register(const std::string& name);
+GALOIS_EXPORT galois::Result<void> Register(const std::string& name);
+
+/// Forget an RDG, detaching it from the namespace
+/// \param name is storage location prefix that the RDG is stored in
+GALOIS_EXPORT galois::Result<void> Forget(const std::string& name);
```
ddn0, line 300: just checking: we call this forget instead of delete because deletion will happen (if it happens at all) during a separate garbage collection pass?

ddn0 (approved):
I'm looking forward to the CI tests that use all of this functionality :D

libtsuba/include/tsuba/RDG.h
tylershunt: yeah, I expect files to be removed from storage by `tsuba_gc`.

libsupport/include/galois/Http.h
```diff
@@ -10,44 +10,64 @@ GALOIS_EXPORT Result<void> HttpInit();
...
+GALOIS_EXPORT Result<void> HttpDelete(
+    const std::string& url, std::vector<char>* response);
+
+template <typename T, typename Callable, typename... Args>
 Result<T>
-HttpGetJson(const std::string& url) {
-  std::vector<char> buffer;
-  if (auto res = HttpGet(url, &buffer); !res) {
+HttpOpJson(Callable func, Args&&... args) {
+  std::vector<char> response;
```
witchel, line 32: Yummy parameter pack.

libsupport/src/Http.cpp
```diff
@@ -2,163 +2,181 @@
...
+    int64_t response_code;
+    curl_easy_getinfo(handle_, CURLINFO_RESPONSE_CODE, &response_code);
+    switch (response_code) {
+    case 200:
+      return galois::ResultSuccess();
+    case 404:
+      return galois::ErrorCode::NotFound;
+    case 400:
+      return galois::ErrorCode::InvalidArgument;
+    default:
```
witchel, line 103: We lean on `ErrorCode::InvalidArgument` a lot.  Probably not worth addressing at this point, but I do notice it.

libtsuba/include/tsuba/RDG.h
witchel: Also just checking, but if I create a graph `g`, forget it, then create another `g` how are names of the meta files managed so they don't conflict?

libtsuba/tests/tsuba_unlink.cpp
```diff
@@ -0,0 +1,89 @@
+#include <unistd.h>
```
witchel, line 1: Should we call this` tsuba_forget` to emphasize the difference between forget and unlink?

lonestar/analytics/distributed/pagerank/pagerank_push.cpp
```diff
@@ -610,5 +610,8 @@ main(int argc, char** argv) {
         outputLocation, results.data(), globalIDs.data(), results.size());
   }
 
+  if (!write_res) {
+    return 1;
+  }
   return 0;
```
witchel, line 616: Thanks!

witchel (approved):
Looking good.

libtsuba/include/tsuba/RDG.h
tylershunt: Graphs are named by their directory so in that case the meta files would conflict.

libsupport/src/Http.cpp
tylershunt: Agree with this point in general; the description for http error 400 is "Bad Request" and most descriptions talk about improper syntax so it does seem apt here.

PR #541, opened 2020-10-07
Case support

bowu:
This commit supports case expressions without NULL.


libquery/include/querying/Column.h
```diff
@@ -14,6 +14,23 @@ namespace galois {
 using ExpressionValue = std::variant<
     bool, std::string, uint32_t, int32_t, uint64_t, int64_t, float, double>;
 
+// For comparing a variant with a value of its alternative type
+template <typename T, class... Types>
+inline bool
+operator==(const T& t, const std::variant<Types...>& v) {
```
ddn0, line 20: why not use https://en.cppreference.com/w/cpp/utility/variant/operator_cmp e.g.,
```cpp
return a == std::variant<Types>(b);
```

Also it doesn't look like this code is used at all.

libquery/include/querying/Expression.h
```diff
@@ -150,7 +153,8 @@ class Expression {
   std::string ToString();
 
   //! Return the literal that results from evaluation of the arithmetic tree
-  Column Evaluate(const Table* table, const graphs::AttributedGraph* graph);
+  virtual Column Evaluate(
```
ddn0, line 156: can you replace `virtual` with `override` (at the end of the declaration)?

libquery/src/CaseExpression.cpp
```diff
@@ -0,0 +1,110 @@
...
+  std::vector<Column> case_cols;
+  std::vector<Column> alternative_cols;
+  for (auto& kv : case_pairs_) {
+    case_cols.push_back(kv.predicate->Evaluate(table, graph));
+    alternative_cols.push_back(kv.result->Evaluate(table, graph));
+  }
+  Column default_col = default_exp_->Evaluate(table, graph);
+
+  std::vector<int> ind_array;
+  switch (val_col.GetType()) {
```
ddn0, line 65: Looks like this will conflict with #540

libquery/src/GroupByOp.cpp
```diff
@@ -103,99 +103,99 @@ galois::GroupByOp::LocalAggregation(
...
-    std::vector<Column> tmp_cols(items_.size());
-    std::vector<ExpressionValue> first_row(items_.size());
-    int i = 0;
-    for (auto& col : key_cols) {
-      tmp_cols[key_inds[i]] = col;
-      i++;
+  std::vector<Column> tmp_cols(items_.size());
+  std::vector<ExpressionValue> first_row(items_.size());
+  int i = 0;
+  for (auto& col : key_cols) {
```
ddn0, line 144: `const auto&`

libquery/src/GroupByOp.cpp
```diff
@@ -103,99 +103,99 @@ galois::GroupByOp::LocalAggregation(
...
-      auto selection = arrow::compute::Take(*col.GetData(), *ind_array);
-      GALOIS_LOG_ASSERT(selection.ok());
-      perm_cols.emplace_back(
-          Column(col.GetType(), std::move(selection.ValueOrDie())));
-    }
+  // Permute all rows based on the index array
+  std::vector<Column> perm_cols;
+  for (const auto& col : used_cols) {
+    auto selection = arrow::compute::Take(*col.GetData(), *ind_array);
+    GALOIS_LOG_ASSERT(selection.ok());
```
ddn0, line 126: this assertion seems unnecessary if we are going to use `selection.ValueOrDie`

libquery/src/GroupByOp.cpp
```diff
@@ -103,99 +103,99 @@ galois::GroupByOp::LocalAggregation(
...
-    int i = 0;
-    for (auto& col : key_cols) {
-      tmp_cols[key_inds[i]] = col;
-      i++;
+  std::vector<Column> tmp_cols(items_.size());
+  std::vector<ExpressionValue> first_row(items_.size());
+  int i = 0;
+  for (auto& col : key_cols) {
+    tmp_cols[key_inds[i]] = col;
+    i++;
```
ddn0, line 146: nit: pre-increment (my preference) or post-increment consistently

libquery/src/GroupByOp.cpp
```diff
@@ -103,99 +103,99 @@ galois::GroupByOp::LocalAggregation(
...
+  }
+  size_t accum = 0;
+  Table::Schema output_schema;
+  // Evaluate one row to know the type of each value column
+  // We have this hack because we don't know the type of the expression
+  for (const auto& key_val : map) {
```
ddn0, line 152: no need to fix this, but the idea of getting the first row is probably a good candidate to factor out into a function.

This whole function of `LocalAggregation` is pretty long as it is. Long bodies, especially long loop bodies, are hard to maintain.

ddn0 (approved):
cc @danielmawhirter another use of Table to fix for you :D

libquery/include/querying/Expression.h
bowu: I'm a bit confused. Expression is the base class of CaseExpression.

libquery/src/GroupByOp.cpp
bowu: I need to work with Daniel to make GroupBy nicer, and it will be another PR.

libquery/include/querying/Column.h
bowu: I intended to use it but finally didn't. I'll remove it.

libquery/include/querying/Expression.h
ddn0: I see. It's a little odd to me that Expression isn't abstract, but it should be fixed latter.

Subclasses of expression will inherit things they do not need or want which is usually a sign that the interface is inappropriate, and subclasses also might not reimplement `Evaluate` which is also probably a bug for any concrete subclass of `Expression`.

bowu: I totally agree. As we discussed before, the appropriate way is to build an Expression class hierarchy, but we don't have such a need at lest in the near future.

PR #543, opened 2020-10-08
Disable full-table cypher querying tests

roshandathathri:
These currently consume a lot of memory. We'll renable this once
we can reduce the memory usage.


danielmawhirter (approved):
LGTM as a band-aid

roshandathathri:
Closing as this will be fixed in #540

PR #546, opened 2020-10-09
name-service: add docker deployment

tylershunt:
I added a docker deployment scripts that should look familiar (credit to @ddn0 since I ripped off the `demosite` scripts).

`./deploy/docker.sh` launches an `name-server` and a local etcd cluster with 3 nodes, then tells you what IP  and port to point  your katana instance at.

still WIP because I was holding off on the CI integration until we decided what we were going to do with the broader CI direction but you can expect that part soon! I figured I would throw this bit up for comments.


name-service/deploy/docker.sh
```diff
@@ -0,0 +1,103 @@
...
+    echo "Killing containers: ${containers}"
+    docker kill ${containers} &>/dev/null
+  fi
+}
+
+# Containers named
+trap "{ kill_containers 'kns_*'; exit 255; }" EXIT
+kill_containers 'kns_*'
+
+readonly BASE=$(cd $(dirname "$0"); pwd)
```
ddn0, line 22: TIL `readonly`

name-service/deploy/docker.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# deploy Katana Name Service (KNS)
```
ddn0, line 3: FWIW, I'm happy to not use bash scripts for everything. There just happens to be n+1 ways to babysit a set of processes and I figured this would be the easiest for people to follow and repurpose.

name-service/name-server/name-server.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+  "strings"
```
ddn0, line 10: gofmt

name-service/deploy/docker/nameserver
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.15
```
ddn0, line 1: not necessary, but in case you wanted to make a minimal deployed image you could do a multistage build.

I'll probably get the exact syntax wrong but something like:

```dockerfile
FROM golang:1.15 as build

WORKDIR /go/src/name-server

COPY ./name-server .
RUN go get -v ./...
RUN go install -v ./...

FROM debian:XY-slim

COPY --from=build /go/bin/nameserver /usr/local/bin/nameserver
CMD /usr/local/bin/nameserver
```

ddn0 (approved):

name-service/deploy/docker.sh
witchel: Yeah, I looked at https://amoffat.github.io/sh/ one day while daydreaming about replacing all bash scripts with python.

tylershunt: `sh` is great! I used it in `./scripts/cli_ctl`

tylershunt: I ended up just re-writing it in python, lmk what you think!

witchel: f strings, real programming language, what's not to like?  Maybe job control is a little clunkier, but this will be much easier to maintain.

I didn't even know about click.  Looks sharp.

tylershunt:
I'm still debugging CI (if anyone has a better suggestion for CI based service management than `tmux` lmk, but marking this as ready to communicate my intent to merge soon :D.

name-service/deploy/docker.sh
tylershunt: click is the best available python cli tool/arg parser imo.

witchel (approved):
Operational maturity, here we come.

ddn0:
> if anyone has a better suggestion for CI based service management than tmux

Oh there are tons of these things, but in the absence of any other requirements, something that is easy to understand and replace if needed like `tmux` is probably the best balance to strike.

The next inflection point will be when we pick a packaging format (deb, docker, VM images, etc) and the CI does a test deployment using those artifacts.

PR #549, opened 2020-10-13
Download new inputs if missing

ddn0:
A common test failure is simply not updating the test inputs in the
build directory if a new version of the inputs is available.  This
commit automatically updates test inputs in the build directory to match
the version indicated in the source directory.

Some care has been taken to not download test inputs in cases where no
testing is expected, e.g., BUILD_TESTING=OFF or an explicit input
location is given at configuration time. However, the default build
configuration will now attempt to download test inputs which may
surprise some unfamiliar with the codebase.

In the short term, it seems beneficial to reduce the number of
configuration steps and we can consider more fine-grain control if there is
more demand for it later.


witchel:
I think this is a very useful change, thank you.  Most development environments can handle the 125MB of inputs I think.

tylershunt (approved):
👨‍🍳 LGTM

gurbinder533 (approved):
This is great.

